PR #1, opened 2022-07-14
DM-35591 Prototype AnalysisTools

natelust:


python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
```diff
@@ -0,0 +1,177 @@
...
+    KeyedDataSchema,
+    Vector,
+)
+from ..scalarActions import CountAction, MedianAction, SigmaMadAction
+from ..keyedDataActions import AddComputedVector, KeyedScalars
+from ..vectorActions import SnSelector, StellarSelector, FractionalDifference
+from ..vectorActions.calcShapeSize import CalcShapeSize
+
+
+class _AproxMedian(ScalarAction):
```
bechtol, line 28: Should this class be moved to scalarActions/scalarActions.py ?

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
```diff
@@ -0,0 +1,177 @@
...
+    vectorKey = Field[str](doc="Key for the vector to perform action on", optional=False)
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return ((self.vectorKey, Vector),)  # type: ignore
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        mask = self.getMask(**kwargs)
+        value = np.sort(data[self.vectorKey.format(**kwargs)][mask])  # type: ignore
+        x = int(len(value) / 10)
+        return np.nanmedian(value[-x:])
```
bechtol, line 38: I think this should be 

return np.nanmedian(value[::x])

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
laurenam: Also, is "Aprox" here meant to be short for "Approximately"?  If so, check your "p"s (and if not, perhaps a full spell-out would avoid confusion?)

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
```diff
@@ -0,0 +1,177 @@
...
+    snFluxType = Field[str](doc="column key for the flux type used in SN selection")
+
+    selector = ConfigurableActionField[VectorAction](doc="Selector to use before computing Scalars")
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.scalarActions.median = MedianAction(colKey=self.columnKey)  # type: ignore
+        self.scalarActions.sigmaMad = SigmaMadAction(colKey=self.columnKey)  # type: ignore
+        self.scalarActions.count = CountAction(colKey=self.columnKey)  # type: ignore
+        self.scalarActions.approxMag = _AproxMedian(  # type: ignore
```
laurenam, line 53: Mind your "p"s (but worry not about any "q"s) 😉

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
natelust: This was intended to be something specific implementation detail specific for how this calculates something, and may get replaces with something better later. I have no strong feeling if others really want to move it I will not argue, but if others don't have strong feelings I would prefer to leave it where it is.

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
natelust: No, in this case it is doing exactly what it should. it sorts the array, finds points that are considered the "tail" and only wants to find the median of those tail values.

This is used as a heuristic to determine approximately what magnitude corresponds to a certain signal to noise.

python/lsst/analysis/tools/scalarActions/scalarActions.py
```diff
@@ -0,0 +1,79 @@
...
+    vectorKey = Field[str](doc="Key for the vector to perform action on", optional=False)
+
+    def getInputSchema(self, **kwargs) -> KeyedDataSchema:
+        return ((self.vectorKey.format(**kwargs), Vector),)
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        mask = self.getMask(**kwargs)
+        value = np.sort(data[self.vectorKey.format(**kwargs)][mask])  # type: ignore
+        x = int(len(value) / 10)
+        return np.nanmedian(value[-x:])
```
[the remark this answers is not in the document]
natelust: ah in contrast to what was above, I recall talking to Sophie about her wanting it else where so perhaps we need to consolidate versions. See other comments for discussion on internals

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
```diff
@@ -0,0 +1,278 @@
...
+        self.prep.selectors.starSelector = StellarSelector()
+
+        self.process.buildActions.x = ExtinctionCorrectedMagDiff()
+        self.process.buildActions.x.magDiff.returnMillimags = False
+        self.process.buildActions.y = ExtinctionCorrectedMagDiff()
+        self.process.buildActions.y.magDiff.returnMillimags = False
+
+
+class GRIStellarPSFMetric(AnalysisMetric):
+    # Use this as the Base Class for now StellarLocusBaseMetric
```
bechtol, line 94: This comment seems inconsistent with the class inheritance pattern that is being used.

python/lsst/analysis/tools/analysisParts/shapeSizeFractional.py
bechtol: Oh, I was thrown off by the name of the class suggesting a median

PR #8, opened 2022-07-21
DM-35610: Add some yamls to the pipelines directory that can be run in ci_imsim

yalsayyad:


ups/analysis_tools.table
```diff
@@ -3,6 +3,7 @@
 # - Common third-party packages can be assumed to be recursively included by
 #   the "base" package.
 setupRequired(base)
+setupRequired(pipe_tasks)
```
natelust, line 6: Since you are keeping up with this anyway, can you add a required for pipe_base and daf_butler ?

natelust (approved):

PR #9, opened 2022-07-21
DM-35617: Add photometric repeatability metric

bechtol:


python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -272,3 +277,71 @@ def getInputSchema(self) -> KeyedDataSchema:
...
+    selectors = ConfigurableActionStructField[VectorAction](
+        doc="Selectors for selecting rows, will be AND together",
+    )
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        for action in self.selectors:
+            yield from action.getInputSchema()
+
+    def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
+        mask: Optional[np.ndarray] = None
```
natelust, line 293: if possible use `Vector` instead of `np.ndarray`, as it will be more portable whenever I get to making a proper Protocol for it

natelust: Also, I have not been consistent in other places, I think I only did it in `SacalarActions` but I had a thought that possible someone might pass `mask=<some_mask>` in kwargs, possibly some extra mask that has been defined somewhere, like a notebook or something. Do you think it is worth it to change this to `mask: Vector | None = kwargs.get('mask')` ?

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -272,3 +277,71 @@ def getInputSchema(self) -> KeyedDataSchema:
...
+        return cast(Vector, mask)
+
+
+class ThresholdSelector(VectorAction):
+    """Return a mask corresponding to an applied threshold.
+    """
+
+    op = Field[str](doc="Operator name.")
+    threshold = Field[float](doc="Threshold to apply.")
+    columnKey = Field[str](doc="Name of column")
```
natelust, line 309: I know this is terribly inconsistent because of changing interfaces in development, but can you change this to `VectorKey` here and anywhere else you see it (though changing in other places needs tests to make sure we dont break things). I am trying to get away from the idea that these are columns in a table. Changing it at least here will keep the name from spreading further.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -272,3 +277,71 @@ def getInputSchema(self) -> KeyedDataSchema:
...
+
+    def __call__(self, data: KeyedData, **kwargs) -> Vector:
+        mask = getattr(operator, self.op)(data[self.columnKey], self.threshold)
+        return cast(Vector, mask)
+
+
+class BandSelector(VectorAction):
+    """Makes a mask for sources observed in a specified set of bands."""
+
+    columnKey = Field[str](
```
natelust, line 322: same as above

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -200,3 +231,27 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+    """Compute per-group statistic values."""
+
+    groupKey = Field[str](doc="Column key to use for forming groups", default="obj_index")
+    buildAction = ConfigurableActionField(doc="Action to build vector", default=LoadVector)
+    func = Field[str](doc="Name of function to be applied per group", default="count")
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return tuple(self.buildAction.getInputSchema()) + ((self.groupKey, Vector),)
+        #yield from self.buildAction.getInputSchema()
+        #yield (self.groupKey, Vector)
```
natelust, line 246: code fragment

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -200,3 +231,27 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+
+    def __call__(self, data: KeyedData, **kwargs) -> Vector:
+
```
natelust, line 249: odd spacing

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
```diff
@@ -291,3 +296,49 @@ def setDefaults(self):
             "stdevSky": "nJy",
             "sigmaMADSky": "nJy",
         }
+
+
+class StellarPhotometricRepeatabilityMetric(AnalysisMetric):
+    parameterizedBand: bool = True
```
natelust, line 302: Don't re-define the existing default

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
```diff
@@ -291,3 +296,49 @@ def setDefaults(self):
...
+    fluxType: str = "psfFlux"
```
natelust, line 303: Are you planning on subclassing in some way? Why have this class level variable?

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
```diff
@@ -291,3 +296,49 @@ def setDefaults(self):
...
+            columnKey="perGroupExtendedness", op="le", threshold=0.5,
+        )
+
+        self.process.calculateActions.photRepeatStdev = MedianAction(colKey="perGroupStdev")
+
+        self.produce.units = {  # type: ignore
+            "photRepeatStdev": "mag",
+        }
+        self.produce.newNames = {
+            "photRepeatStdev": "{band}_stellarPhotRepeatStdev"
```
natelust, line 343: I see `{band}` here, but I dont see band elsewhere and you are defining psfFlux not per band. where does the per-band calculation come int?

python/lsst/analysis/tools/actions/vector/vectorActions.py
bechtol: Good catch. It looks like that was a remnant of unfinished thought.

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
bechtol: I wanted to ensure that the same fluxType is used throughout the calculation. Is there a better way to do that?

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -20,39 +20,36 @@
...
-# from ..analysisMetrics.analysisMetrics import ShapeSizeFractionalMetric
-
 
 class AssociatedSourcesTractAnalysisConnections(
     AnalysisBaseConnections,
     dimensions=("skymap", "tract"),
     defaultTemplates={
-        "inputName": "isolated_star_sources",
-        # "associatedSourcesInputName": "isolated_star_sources"},
+        "outputName": "isolated_star_sources",
```
jeffcarlin, line 35: Interesting that you included a default for `outputName`. I don't think we've been doing this elsewhere in `analysis_tools`, but it seems like good practice.

python/lsst/analysis/tools/analysisMetrics/photometricRepeatabilityMetrics.py
```diff
@@ -0,0 +1,106 @@
...
+        self.process.calculateActions.photRepeatStdev = MedianAction(vectorKey="perGroupStdevFiltered")
+        self.process.calculateActions.photRepeatOutlier = FracThreshold(
+            vectorKey="perGroupStdevFiltered",
+            op="ge",
+            threshold=0.015,
+            percent=True,
+        )
+
+        self.produce.units = {  # type: ignore
+            "photRepeatStdev": "mag",
```
jeffcarlin, line 100: It would make more sense for the units to be mmag by default?

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
```diff
@@ -183,3 +189,60 @@ def setDefaults(self):
...
+        )
+        self.process.buildActions.perGroupStdev.func = "std"
+
+        # Filter on per-group quantities
+        self.process.filterActions.perGroupStdevFiltered = DownselectVector(vectorKey="perGroupStdev")
+        self.process.filterActions.perGroupStdevFiltered.selectors.count = ThresholdSelector(
+            vectorKey="perGroupCount",
+            op="ge",
+            threshold=3,
+        )
```
jeffcarlin, line 224: When I tried to execute the plotting portion of the code, it gave an error on line 220 that "AttributeError: 'DownselectVector' object has no attribute 'selectors'". I think you either need to create an empty dict (or list?) before that line, or something similar to give it a "selectors" attribute.

jeffcarlin (approved):
Overall I think it looks good! I did a test run on rc2_subset, and it ran fine.

There are a few comments, but otherwise the main barrier to merging is making sure the plot works. (I tried to test that, but came up against some errors.)

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
bechtol: I am removing the plotting capability from this PR.

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
```diff
@@ -183,3 +189,60 @@ def setDefaults(self):
...
+
+        self.process.calculateActions.stars = ScatterPlotStatsAction(
+            vectorKey="yStars",
+        )
+
+        self.produce = ScatterPlotWithTwoHists()
+
+        self.produce.plotTypes = ["stars"]
+        self.produce.xAxisLabel = "PSF Magnitude (mag)"
+        self.produce.yAxisLabel = "Fractional size residuals (S/S_PSF - 1)"
```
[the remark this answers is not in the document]
bechtol: Removing plotting capability from this PR

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
```diff
@@ -183,3 +189,60 @@ def setDefaults(self):
...
+        self.process.filterActions.perGroupStdevFiltered.selectors.sn = ThresholdSelector(
+            vectorKey="perGroupSn",
+            op="ge",
+            threshold=200,
+        )
+        self.process.filterActions.perGroupStdevFiltered.selectors.extendedness = ThresholdSelector(
+            vectorKey="perGroupExtendedness",
+            op="le",
+            threshold=0.5,
+        )
```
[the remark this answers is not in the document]
bechtol: Removing plotting capability from this PR.

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
```diff
@@ -219,3 +220,25 @@ def setDefaults(self):
...
+        self.process.calculateActions.meanSky = MeanAction(vectorKey=f"{{band}}_{self.fluxType}")
+        self.process.calculateActions.stdevSky = StdevAction(vectorKey=f"{{band}}_{self.fluxType}")
+        self.process.calculateActions.sigmaMADSky = SigmaMadAction(vectorKey=f"{{band}}_{self.fluxType}")
+
+        self.produce.units = {  # type: ignore
+            "medianSky": "nJy",
+            "meanSky": "nJy",
+            "stdevSky": "nJy",
+            "sigmaMADSky": "nJy",
+        }
```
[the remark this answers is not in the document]
bechtol: Good catch. I think I might have included this accidentally when doing an interactive a rebase.

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -200,3 +258,19 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
             correction = correction.to(u.mmag)
 
         return np.array(diff - correction.value)
+
+
+class PerGroupStatistic(VectorAction):
+    """Compute per-group statistic values."""
+
```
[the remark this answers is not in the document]
bechtol: Expanding the documentation as suggested.

python/lsst/analysis/tools/actions/keyedData/keyedDataActions.py
```diff
@@ -123,8 +123,9 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
                 mask = subMask
             else:
                 mask *= subMask  # type: ignore
-
-        result = {key.format_map(kwargs): data[key.format_map(kwargs)] for key in self.vectorKeys}
+        result = {}
+        for key in self.vectorKeys:
+            result[key.format_map(kwargs)] = data[key.format_map(kwargs)]
```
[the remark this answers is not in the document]
bechtol: I think this is a new change. The previous version was throwing an error. As I understand, python dictionary comprehensions have their own scope (see, e.g., [here](https://stackoverflow.com/questions/62208744/pythondict-comprehension-and-eval-function-variable-scope)). In this case, I thought the code would be more readily understandable by avoiding a dictionary comprehension.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -79,3 +80,25 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
         value = np.sort(data[self.vectorKey.format(**kwargs)][mask])  # type: ignore
         x = int(len(value) / 10)
         return np.nanmedian(value[-x:])
+
+
+class FracThreshold(ScalarAction):
+    """Compute the fraction of a distribution of values relative to a threshold."""
+
```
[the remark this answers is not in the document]
bechtol: Expanding the docstring to provide examples and making the operator parameter a ChoiceField.

python/lsst/analysis/tools/analysisMetrics/analysisMetrics.py
bechtol: We are performing measurements on sources catalogs which have `psfFlux` measurements. Each source is associated with an exposure which is turn is associated with a band, such that each source has a `band` column. In the prep stage, we select only the source measurements in the band of interest.

`self.prep.selectors.bandSelector = BandSelector()`

python/lsst/analysis/tools/analysisMetrics/photometricRepeatabilityMetrics.py
bechtol: Indeed, the specifications are stated in mmag, so converting to use mmag

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -52,6 +53,33 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+        mask: Optional[Vector] = None
+        for selector in self.selectors:
+            subMask = selector(data, **kwargs)
+            if mask is None:
+                mask = subMask
+            else:
+                mask *= subMask  # type: ignore
+        return cast(Vector, data[self.vectorKey.format(**kwargs)])[mask]
+
+
```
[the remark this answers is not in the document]
bechtol: The `DownselectVector` returns a vector where a single configurable action has been used to apply selection criteria. Usually this will be a shorter vector due to the applied filtering.

In this instance, I needed to apply several per-group selection criteria (e.g., number of repeated measurements, mean signal-to-noise ratio) to keep only the per-group signal-to-noise measurements of interest. 

The first time I implemented this, I introduced the `AndSelector`, but then I felt that this made the syntax more complicated/verbose than I wanted. Basically I needed to introduce the new AndSelector and then set the multiple selectors for that selector. It added an extra layer of nesting that seemed unwarranted. 

I also considered to make `DownselectVector` always allow for the possibility of multiple selectors by using `selectors = ConfigurableActionStructField` rather than `selector = ConfigurableActionField`. This also seemed unsatisfying because one would need to add an extra layer in cases where there is only a single criterion applied. For example

```
self.process.filterActions.xStars = DownselectVector(
    vectorKey="mags", selector=VectorSelector(vectorKey="starSelector")
)
```

would become

```
self.process.filterActions.xStars = DownselectVector(vectorKey="mags")
self.process.filterActions.xStars.selectors.star = VectorSelector(vectorKey="starSelector")
```

I wasn't sure that we wanted to introduce that additional layer.

I'm still not fully convinced that introducing `MultiCriteriaDownselectVector` is the optimal solution.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -272,3 +277,71 @@ def getInputSchema(self) -> KeyedDataSchema:
 
     def __call__(self, data: KeyedData, **kwargs) -> Vector:
         return cast(Vector, data[self.vectorKey.format(**kwargs)])
+
+
+class AndSelector(VectorAction):
```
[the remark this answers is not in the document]
bechtol: I would propose not to add the `AndSelector` and instead use `MultiCriteriaDownselectVector`.

python/lsst/analysis/tools/actions/vector/selectors.py
bechtol: This got fixed along the way.

python/lsst/analysis/tools/actions/vector/vectorActions.py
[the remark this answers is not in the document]
natelust: I might be remembering the And selector you had, but I could have sworn there was something different elsewhere. Anyway, the way I had imagined this to work is that you would build the mask you want (with all the various selectors) in the build stage, and then you would just apply that mask in the filter stage, and thus you would only need the one selector (`VectorSelector`), and then you would have the vector around for anything else you might need it for. I am not saying things MUST be done how I thought them out, just that that process made sense to me reading things out. I wanted to bring that up to the discussion for consideration, but if you determine your way is best I am happy to have the new selector, but it does make the other kind of redundant.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -325,3 +330,70 @@ def getInputSchema(self) -> KeyedDataSchema:
 
     def __call__(self, data: KeyedData, **kwargs) -> Vector:
         return cast(Vector, data[self.vectorKey.format(**kwargs)])
+
+
+# class AndSelector(VectorAction):
```
natelust, line 335: If you don't intend this to go in (see later comment) it should be deleted

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -82,6 +115,36 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+        Returns
+        -------
+        result : `Vector`
+            Computed signal-to-noise ratio.
+        """
+        fluxCol = self.fluxType.format(**(kwargs | dict(band=self.band)))
+        errCol = f"{fluxCol}{self.uncertaintySuffix.format(**kwargs)}"
+        result = cast(Vector, data[fluxCol]) / data[errCol]  # type: ignore
+
+        return np.array(cast(Vector, result))
```
natelust, line 145: Was it you I discussed this with before and it was resolved? (Sorry I have been looking at so many reviews recently)

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -325,3 +330,70 @@ def getInputSchema(self) -> KeyedDataSchema:
...
+    vectorKey = Field[str](doc="Key of the Vector which defines the band", default="band")
+    bands = ListField[str](
+        doc="The bands to select. `None` indicates no band selection applied.",
+        default=[],
+    )
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return ((self.vectorKey, Vector),)
+
+    def __call__(self, data: KeyedData, **kwargs) -> Vector:
```
natelust, line 384: MyPy is basically saying it cant work out what type you want bands to be, because match arms are defining it to be different things:
1) is defining it as tuple[Any, ...] - a multi element tuple of any type
2) I think it is infering the same as 1
3) it thinks it is a tuple with a single element of type Any i.e. tuple[Any]
4) None

solution to all this is before the match statement add:
`bands: tuple[Any, ...] | None`
That declares the type for the variable before it is used anywhere

PR #11, opened 2022-07-21
DM-35676: Fix the broken stellar locus plot in analysis_tools

sr525:


python/lsst/analysis/tools/interfaces.py
```diff
@@ -420,6 +420,9 @@ def getOutputNames(self) -> Iterable[str]:
         """
         outNames = tuple(self.produce.getOutputNames())
         if outNames:
-            return outNames
+            return (f"{self.identity or ''}_{name}" for name in outNames)
```
leeskelvin, line 423: If `self.identity` is `None`, then the return here will be `_{name}`, which I don't think was the desired intention. This could be changed to be smarter about naming things, but I'm not going to push for it on this ticket.

python/lsst/analysis/tools/interfaces.py
```diff
@@ -420,6 +420,9 @@ def getOutputNames(self) -> Iterable[str]:
...
         else:
-            return (f"{{band}}_{self.identity or ''}",)
+            if self.parameterizedBand:
+                return (f"{{band}}_{self.identity or ''}",)
```
leeskelvin, line 426: As above, this could end up returning `{band}_`, but, happy to ignore this for the time being.

python/lsst/analysis/tools/actions/plot/colorColorFitPlot.py
```diff
@@ -240,29 +244,29 @@ def makePlot(
...
+            xsFitLine = (ysFitLine - data["bODR"]) / data["mODR"]
+            ysFitLine2 = np.array([data["yMin"], data["yMax"]])
+            xsFitLine2 = (ysFitLine2 - data["bODR2"]) / data["mODR2"]
 
         else:
-            xsFitLineHW = np.array([fitParams["xMin"], fitParams["xMax"]])
-            ysFitLineHW = fitParams["mHW"] * xsFitLineHW + fitParams["bHW"]
-            xsFitLine = np.array([fitParams["xMin"], fitParams["xMax"]])
+            xsFitLineHW = np.array([data["xMin"], data["xMax"]])
+            ysFitLineHW = data["mHW"] * xsFitLineHW + data["bHW"] # type: ignore
```
leeskelvin, line 257: This needs a double space pre-comment.

PR #13, opened 2022-07-22
Dm-35650: Handle multiple inputs in reconstruction

natelust:
The reconstructAnalysisTools function should handle when inputs
are expected to be multiple.

### DIFF
--- a/python/lsst/analysis/tools/tasks/reconstructor.py
+++ b/python/lsst/analysis/tools/tasks/reconstructor.py
@@ -20,27 +20,78 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any, Callable
 
-from lsst.daf.butler import Butler, DataId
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/tasks/reconstructor.py
```diff
@@ -20,27 +20,78 @@
...
+    from lsst.daf.butler import Butler, DataId
+
 
 def reconstructAnalysisTools(
-    butler: Butler, collection: str, label: str, dataId: DataId
+    butler: Butler,
+    collection: str,
+    label: str,
+    dataId: DataId,
+    callback: Callable[[dict[str, Any], DataId], dict[str, Any]] | None,
```
bechtol, line 39: We might want the callback to default to `None`. Otherwise, the presence of the required argument might be confusing for situations when no callback function is needed.

bechtol (approved):
See the comment about callback argument having a default.

I have verified the basic functionality, but have not tested the callback functionality yet because this requires developing a callback function with the right argument signature.

PR #15, opened 2022-07-22
DM-35619: Add astrometry residuals with refCat plots

cmsaunders:


python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -0,0 +1,408 @@
...
+            targetMatchIndices: `numpy.ndarray`
+                Array of indices of matched target catalog objects
+            separations: `astropy.coordinates.angles.Angle`
+                Array of angle separations between matched objects
+        """
+        refCat_ap = SkyCoord(ra=refCatalog['coord_ra'] * u.degree,
+                             dec=refCatalog['coord_dec'] * u.degree)
+
+        sourceCat_ap = SkyCoord(ra=targetCatalog['coord_ra'] * u.degree,
+                                dec=targetCatalog['coord_dec'] * u.degree)
```
sr525: Should we have spaces around the * here?

sr525: Do we know that the units are always gong to be degrees or should it be a config option?

python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -0,0 +1,408 @@
...
+
+    skymap = pipeBase.connectionTypes.Input(
+        doc="The skymap for the tract",
+        storageClass="SkyMap",
+        name=BaseSkyMap.SKYMAP_DATASET_TYPE_NAME,
+        dimensions=("skymap",)
+    )
+
+    matchCatalog = pipeBase.connectionTypes.Output(
+        doc="Catalog with matched target and reference objects with separations",
```
sr525, line 132: I think I would prefer matchedCatalog as it reinforces that this is the one that has been matched rather than the one to match but I don't think it really matters.

python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -0,0 +1,408 @@
...
+
+        inputs = butlerQC.get(inputRefs)
+
+        columns = ['coord_ra', 'coord_dec', 'patch'] + self.config.extraColumns.list()
+        for selectorAction in [self.config.selectorActions, self.config.sourceSelectorActions,
+                               self.config.extraColumnSelectors]:
+            for selector in selectorAction:
+                for band in ["g", "r", "i", "z", "y"]:
+                    selectorSchema = selector.getFormattedInputSchema(band=band)
+                    columns += [s[0] for s in selectorSchema]
```
sr525, line 211: Mixture of ' and ". I don't think it matters but it looks neater when it is consistent.

python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -0,0 +1,408 @@
...
+        epoch = Time(self.config.epoch, format="decimalyear")
+
+        skyCircle = self.refObjLoader.loadSkyCircle(center,
+                                                    radius,
+                                                    'i',
+                                                    epoch=epoch)
+        refCat = skyCircle.refCat
+
+        # Convert the coordinates to RA/Dec and convert the catalog to a
+        # dataframe
```
sr525, line 299: Spaces around *

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -71,7 +74,10 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
         mask = self.getMask(**kwargs)
         value = np.sort(cast(Vector, data[self.vectorKey.format(**kwargs)])[mask])
         x = int(len(value) / 10)
-        return np.nanmedian(value[-x:])
+        median = np.nanmedian(value[-x:])
+        if self.inputUnit != self.outputUnit:
+            median = (median * u.Unit(self.inputUnit)).to(u.Unit(self.outputUnit)).value
+        return median
```
sr525, line 80: Spaces around *

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -200,3 +200,45 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+
+
+class AstromDiff(VectorAction):
+    """Calculate the difference between two columns, assuming their units
+    are degrees, and convert the difference to arcseconds.
+    Parameters
+    ----------
+    df : `pandas.core.frame.DataFrame`
+        The catalog to calculate the position difference from.
+    Returns
```
sr525, line 212: Variable type and name missing.

sr525:
I've left a few minor comments, I know that you have tested this all thoroughly and that it works. One general comment is that I think the style guideline says that we shouldn't have spaces around * and / but there are spaces around them here. Perhaps black did this formatting?

python/lsst/analysis/tools/tasks/catalogMatch.py
cmsaunders: My mind is blown -- somehow I have been doing spacing around operators wrong. 

I added a config option for the units.

timj: When you switch the code to using `black` you will find that that rule about spaces around operators goes away anyhow...

python/lsst/analysis/tools/tasks/catalogMatch.py
cmsaunders: This was moved to a config option (with consistent quotation marks).

python/lsst/analysis/tools/tasks/catalogMatch.py
natelust: Are you using black, this is a thing that has a right answer, and black does it.

cmsaunders: Yeah, when I tried to commit, triggering `pre-commit`, `black` complained about all the places I just removed whitespaces. However, this seems to disagree with https://developer.lsst.io/python/style.html#binary-operators-should-be-surrounded-by-a-single-space-except-for. Maybe I don't have `black` configured correctly.

timj: When black is enabled for a project the Python style guide for formatting is no longer relevant. I thought there was a comment about that somewhere in the dev guide but maybe that was only in the section about using black.

natelust: https://developer.lsst.io/python/style.html?highlight=black#black-may-be-used-to-fix-pep-8-compliance

cmsaunders:
@timj , do you mean in interfaces.py?

timj:
@cmsaunders  I was commenting on a line in a commit but that clearly became detached from the pull request so sorry about that.

It was the `type: ignore` on the `Field` that I was worrying about. We solved that one in other packages by annotating the variable that the `Field` is being assigned to.

cmsaunders:
I tried `parameterizedBand: Field[bool] =`, but it didn't work how I need it to in the class. I don't know if it would have passed mypy.

natelust:
@timj in this cases @cmsaunders is using something that ducks like a bool on the class interface, but MYPY is unhappy because of LSP even though they duck the same. We could loosen up the type on the base class, but in reality we likely will move toward something similar to what @cmsaunders  has on the base class, we just dont want to disrupt right before the weekly, so as we know they duck the same, for now we are just trying to inform MYPY it ok to violate LSP.

natelust:
@cmsaunders In light of needing yet another cast or typing or something in order to make mypy happy again, let's back up and actually do the type union on the base class. you can change https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/interfaces.py#L287 this to `parameterizedBand: bool | Field[bool]` I think

natelust:
and remove the type ignore where we added it earlier

cmsaunders:
@natelust, I pushed a change before I saw your last message, which worked, but I'll try your suggestion and see how that works.

PR #19, opened 2022-07-25
dm 35622: Implement tests for scalar and vector actions

fred3m:


python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -194,9 +194,10 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
 
         ebv = data[self.ebvCol]
         # Ignore type until a more complete Vector protocol
-        correction = (av1 - av2) * ebv * u.mag  # type: ignore
+        correction = np.array((av1 - av2) * ebv) * u.mag  # type: ignore
 
         if self.magDiff.returnMillimags:
             correction = correction.to(u.mmag)
+        print(correction, diff)
```
natelust, line 201: Remove this

tests/test_actions.py
```diff
@@ -0,0 +1,428 @@
...
+            "E(B-V)": r[::-1],
+            "r_ixx": r**2,
+            "r_iyy": r[::-1] ** 2,
+            "r_ixy": r * r[::-1],
+            "g_iixx": r**2,
+            "g_iiyy": r[::-1] ** 2,
+            "g_iixy": r * r[::-1],
+        }
+
+        self.data = pd.DataFrame.from_dict(data)
```
natelust, line 121: Why did you turn this into a DataFrame? This should be fine as is as a dictionary...

tests/test_actions.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def checkSchema(self, action, truth):
```
natelust, line 123: should this be private?

natelust (approved):

tests/test_actions.py
fred3m: I just wanted to test with a DataFrame. Before I did, there was at least one test that worked with a Dict but not a DataFrame, so it seemed worth including.

PR #22, opened 2022-08-01
Tickets/DM-35607: Get plot information from run quantum

sr525:


python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -230,12 +234,16 @@ def getInputSchema(self) -> KeyedDataSchema:
...
         base.append(("lowSnThreshold", Scalar))
         base.append(("highSnThreshold", Scalar))
+
+        if self.addSummaryPlot:
+            base.append(("patch", Vector))
+
         return base
 
     def __call__(self, data: KeyedData, **kwargs) -> Mapping[str, Figure] | Figure:
+
```
cmsaunders, line 244: I don't think you want to add a blank line here.

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
```diff
@@ -75,11 +78,24 @@ def setDefaults(self):
             vectorKey="{band}_psfFluxErr", selector=VectorSelector(vectorKey="starSelector")
         )
 
+        self.process.filterActions.patch = DownselectVector(
+            vectorKey="patchWhole", selector=VectorSelector(vectorKey="starSelector")
+        )
+
+        self.process.calculateActions.stars = ScatterPlotStatsAction(
```
cmsaunders, line 85: You need to import `ScatterPlotStatsAction`. This is probably a rebase issue, as the import was removed on `main` recently.

python/lsst/analysis/tools/tasks/base.py
```diff
@@ -193,6 +193,12 @@ def runPlots(self, data: KeyedData, **kwargs) -> Struct:
         if "plotInfo" not in kwargs:
             kwargs["plotInfo"] = _StandinPlotInfo()
         for name, action in self.config.plots.items():
+            for selector in action.prep.selectors:
+                try:
```
cmsaunders, line 197: The `AttributeError` is for cases where `selector.threshold` does not exist, right?Instead of the try/except, how about 
`if "threshold" in selector.keys():
    kwargs["plotInfo"]["SN"] = selector.threshold`

python/lsst/analysis/tools/tasks/base.py
```diff
@@ -305,6 +327,7 @@ def loadData(self, handle: DeferredDatasetHandle, names: Iterable[str] | None =
         """
         if names is None:
             names = self.collectInputNames()
+        names.add("patch")
```
cmsaunders, line 330: Why not just put this line in `collectInputNames`? This would also prevent the mypy error that you are getting because not every `Iterable` has an `add` method.

cmsaunders (approved):
This looks good to me. My only major comment is that it would be nice to add the plot information for the other plot types. It's your call on whether to do this, but you might want to poll people on what will be shown at the PCW, since it would be good to have the plot information on anything that is going to be shown to others.

python/lsst/analysis/tools/tasks/base.py
sr525: This was left over from when I was experimenting with something. I've removed it and everything seems to be working.

sr525:
> This looks good to me. My only major comment is that it would be nice to add the plot information for the other plot types. It's your call on whether to do this, but you might want to poll people on what will be shown at the PCW, since it would be good to have the plot information on anything that is going to be shown to others.

I thought that it was applied to all the plot types but my test pipeline appears to have been demolished somewhere in the recent changes and the standard pipelines don't seem to have a variety of plots in them.

PR #23, opened 2022-08-01
DM-35148: Simplify scipy usage in analysis_tools

taranu:


bechtol (approved):
This looks like a useful set of changes to make the computation of median absolute deviation consistent across the package. I've checked that all the relevant instances of scipy.stats seem to have been replaced throughout the package.

PR #31, opened 2022-09-14
DM-36234: Create AP Number of Associated Solar System Objects metric in analysis_tools

ebellm:


python/lsst/analysis/tools/tasks/diaSourceTableVisitAnalysis.py
```diff
@@ -0,0 +1,31 @@
...
+
+from lsst.pipe.base import connectionTypes as ct
+
+from .base import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class DiaSourceTableCcdVisitAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions=("visit", "band"),
+    defaultTemplates={"coaddName": "deep", "fakesType": "fakes_"},
```
sr525, line 11: We're trying to move to SI rather than fakes, siType and si_.

pipelines/apCcdVisitQualityCore.yaml
```diff
@@ -0,0 +1,16 @@
+description: |
+  Tier 1 AP plots and metrics to assess CcdVisit quality
+tasks:
+  analyzeDiaSourceTableCore:
+    class: lsst.analysis.tools.tasks.DiaSourceTableCcdVisitAnalysisTask
+    config:
+      # Add plots and metrics that work when we have some
+      # plots.label: APlotClass
```
sr525, line 8: This commented out line should go as well.

pipelines/apCcdVisitQualityCore.yaml
```diff
@@ -0,0 +1,16 @@
...
+      metrics.numSsObjects: NumSsObjectsMetric
+      connections.outputName: diaSourceTableCore
+      #metrics.skyFluxVisitStatisticMetric: SkyFluxStatisticMetric
+      #metrics.skyFluxVisitStatisticMetric.applyContext: VisitContext
+      python: |
+        from lsst.analysis.tools.analysisPlots import *
+        from lsst.analysis.tools.analysisMetrics import *
+        from lsst.analysis.tools.contexts import *
```
sr525, line 16: You don't need all of these imports unless you are uncommenting some of the lines.

python/lsst/analysis/tools/tasks/diaSourceTableVisitAnalysis.py
ebellm: We use `fakesType` [throughout](https://github.com/search?q=org%3Alsst+fakesType&type=code) AP, including in the upstream data products, so I propose to maintain it here.

sr525: Benching this discussion for now to be readdressed later.

sr525:
fakees -> si discussion aside this looks good to me.

PR #34, opened 2022-09-29
DM-36238: Create AP Number of Dipoles and Number of DIA Sources metrics in analysis_tools.

erinleighh:


python/lsst/analysis/tools/analysisMetrics/apDiaSourceMetrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+class NumDiaSourcesMetric(AnalysisMetric):
+    """Calculate the number of DIA Sources that do not have known
+    bad/quality flags set to true.
+    """
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.process.buildActions.flagSelector = FlagSelector()
```
natelust, line 41: Since you are not doing any differential filtering (i.e. making two different filter vectors like one for stars and one for galaxies) I would move this flag selector to the prep stage. 

1. Normally quality flags are filtered there anyway as they would apply to all objects
2. It will make interactive use of your Analysis tool easier. The result of prep stage will directly be all the columns you care about and if necessary you can poke around at it for anything weird prior to sending it to the process stage to have computations done on it.

This would look something like
```suggestion
self.prep.selectors.flagSelector = FlagSelector()
self.prep.selectors.flagSelector.selectWhenFalse = [
            "base_PixelFlags_flag_bad",
            "base_PixelFlags_flag_suspect",
            "base_PixelFlags_flag_saturatedCenter",
            "base_PixelFlags_flag_interpolated",
            "base_PixelFlags_flag_interpolatedCenter",
            "base_PixelFlags_flag_edge",
        ]
self.process.calculateActions.numDiaSources = CountAction(vectorKey="diaSourceId")
self.produce.units = {"numDiaSources": "ct"}
```

python/lsst/analysis/tools/analysisMetrics/apDiaSourceMetrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # the units for the quantity (count, an astropy quantity)
+        self.produce.units = {"NumDiaSources": "ct"}
+
+
+class NumDipolesMetric(AnalysisMetric):
+    """Calculate the number of dipoles."""
+
+    def setDefaults(self):
+        super().setDefaults()
```
natelust, line 69: Ah, now that I get down here, something like this for the `AnalysisTool` above

natelust (approved):

natelust (approved):
Sorry I forgot to click approve

PR #36, opened 2022-10-03
DM-36437: Create AP Number of DIA Sources All metric and Count Unique scalar action

erinleighh:


python/lsst/analysis/tools/analysisMetrics/apDiaSourceMetrics.py
```diff
@@ -0,0 +1,78 @@
...
+        # build a filter for bad flags
+        self.process.buildActions.flagSelector.selectWhenFalse = [
+            "base_PixelFlags_flag_bad",
+            "base_PixelFlags_flag_suspect",
+            "base_PixelFlags_flag_saturatedCenter",
+            "base_PixelFlags_flag_interpolated",
+            "base_PixelFlags_flag_interpolatedCenter",
+            "base_PixelFlags_flag_edge",
+        ]
+
```
sr525, line 66: Should we have a standard DIA source flag selector?

erinleighh: Yes, I think that would be helpful!

sr525: Do you want to make one as part of this ticket? Or make another ticket?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -74,6 +74,30 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+    """
+
+    vectorKey = Field[str](doc="Name of column.")
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        yield self.vectorKey, Vector
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        values = cast(Vector, data[self.vectorKey])
+        count = len(np.unique(values))
```
bsmartradio, line 97:
```suggestion
    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
        mask = self.getMask(**kwargs)
        values = cast(Vector, data[self.vectorKey.format(**kwargs)])[mask])
        count = len(np.unique(values))

python/lsst/analysis/tools/analysisMetrics/apDiaSourceMetrics.py
erinleighh: Thank you for your patience! I included it in this ticket and just added it...provided I did the rebase correctly! I also incorporated Bri's notes if you'd like to check those out, too.

PR #38, opened 2022-10-25
DM-36716: Fix pyproject.toml

taranu:


arunkannawadi (approved):
mypy seems to be passing! Yay! 🎉

PR #43, opened 2022-11-21
Tickets/DM-36263: Creating Seeing metric and Assoc/Unassoc Sources metric and plots

bsmartradio:
Create Analysis_Tools metric and plots for unassociatedDiasources and min/max/median seeing. Added a new metric, unique, to count unique items in a vector. Create a base class for sources, which creates three vectors containing all sources, associated, and unassociated sources. Added a pipeline which creates the metrics and plots for assoc/unassociated sources.


pipelines/deepCoaddAssocSources.yaml
```diff
@@ -0,0 +1,13 @@
+description: |
```
isullivan, line 1: This file is called `deepCoaddAssocSources.yaml`, but does it really have anything to do with a `deepCoadd`? Should it be called something like `diaObjectAssociatedSources.yaml` or something similar instead?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -124,3 +124,36 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+
+class MaxAction(ScalarAction):
+    vectorKey = Field[str]("Key of Vector to find maximum")
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return ((self.vectorKey, Vector),)
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        mask = self.getMask(**kwargs)
+        return cast(Scalar, float(np.max(cast(Vector, data[self.vectorKey.format(**kwargs)])[mask])))
```
isullivan, line 137: Is the extra `float(...)` necessary here? Isn't that what `np.max()` returns?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -124,3 +124,36 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+
+class MinAction(ScalarAction):
+    vectorKey = Field[str]("Key for the vector to perform action on")
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return ((self.vectorKey, Vector),)
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        mask = self.getMask(**kwargs)
+        return cast(Scalar, float(np.min(cast(Vector, data[self.vectorKey.format(**kwargs)])[mask])))
```
isullivan, line 148: Same comment about `float(...)`

python/lsst/analysis/tools/analysisMetrics/seeingMetric.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from lsst.analysis.tools.actions.scalar import MaxAction, MedianAction, MinAction
+
+from ..interfaces import AnalysisMetric
+
+
+class SeeingStatistics(AnalysisMetric):
```
isullivan, line 27: I think this file should include a `__all__` statement, so that it behaves nicer with the `import *` in `__init__.py`

python/lsst/analysis/tools/analysisMetrics/sourceMetrics.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from lsst.analysis.tools.actions.scalar import CountAction, FracThreshold
+
+from ..analysisParts.baseSources import BaseSources
+from ..interfaces import AnalysisMetric
+
+
```
isullivan, line 27: This file should also include `__all__`

python/lsst/analysis/tools/analysisMetrics/sourceMetrics.py
```diff
@@ -0,0 +1,51 @@
...
+class CountSources(AnalysisMetric, BaseSources):
+    """Counts associated and unassociated sources as
+    well as their percentage of the total sources."""
```
isullivan, line 30: Include in the description what the sources would have been associated with (diaObjects).

python/lsst/analysis/tools/analysisMetrics/sourceMetrics.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.process.calculateActions.associatedPercent = FracThreshold(
```
isullivan, line 35: Is this actually a percent, with a threshold of >1%?

python/lsst/analysis/tools/analysisParts/baseSources.py
```diff
@@ -0,0 +1,33 @@
+from lsst.analysis.tools.actions.scalar import UniqueAction
+from lsst.analysis.tools.actions.vector import DownselectVector, LoadVector, ThresholdSelector, VectorSelector
+from lsst.analysis.tools.interfaces import AnalysisTool
+
+
```
isullivan, line 5: Add `__all__`

python/lsst/analysis/tools/analysisParts/baseSources.py
```diff
@@ -0,0 +1,33 @@
...
+    """Base class for counting associated and unassociated sources."""
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.process.buildActions.loadVectorSources = LoadVector()
+        self.process.buildActions.associatedVectorSelector = ThresholdSelector()
+        self.process.buildActions.unassociatedVectorSelector = ThresholdSelector()
+
+        # assign keys for PSF and AP Flux
```
isullivan, line 16: What do the next 7 lines have to do with PSF and AP flux?

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
```diff
@@ -329,3 +333,21 @@ def setDefaults(self):
         self.produce.panels["panel_flux"] = HistPanel()
         self.produce.panels["panel_flux"].label = "Psf/Ap Ratio"
         self.produce.panels["panel_flux"].hists = dict(fluxRatioMetric="Ratio")
+
+
+class SourcesPlot(AnalysisPlot, BaseSources):
+    """Base class for counting associated and unassociated sources."""
```
isullivan, line 339: I think this needs an updated docstring about the sorts of plots this supports or produces.

python/lsst/analysis/tools/tasks/diaObjectTableAssocAnalysis.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from lsst.pipe.base import connectionTypes as ct
+
+from .base import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
```
isullivan, line 7: Add `__all__`

python/lsst/analysis/tools/tasks/diaObjectTableAssocAnalysis.py
```diff
@@ -0,0 +1,29 @@
...
+class DiaObjectTableAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions=("visit", "band"),
+    defaultTemplates={"coaddName": "deep", "fakesType": "fakes_"},
```
isullivan, line 11: I think the default should be `"fakesType": ""`, and we should configure the metrics pipeline that uses fakes to update the fakesType.

isullivan (approved):
I had a number of questions about naming and documentation. Also, make sure to add `__all__` to each new file that will have classes imported in `__init__.py`.

python/lsst/analysis/tools/analysisParts/baseSources.py
bsmartradio: This was a stray comment from a notebook. I'll remove it.

python/lsst/analysis/tools/tasks/diaObjectTableAssocAnalysis.py
bsmartradio: Changed it to  `"fakesType": ""`

python/lsst/analysis/tools/actions/scalar/scalarActions.py
bsmartradio: Yes it does return a float, but I noticed that in the other existing scalar actions the type is always specified so I was following that convention. np.max can also return a float if the array only contains ints, though, so I could leave it out and it would give either. However, since all of the other scalar actions seem to require a specific type I decided on a float.

python/lsst/analysis/tools/analysisPlots/analysisPlots.py
bsmartradio: Docstring updated to be more descriptive

PR #44, opened 2022-11-23
DM-37074: Add sky obj hist plots into coaddQualityCore

leeskelvin:


pipelines/coaddQualityCore.yaml
```diff
@@ -26,6 +26,7 @@ tasks:
       metrics.yPerpPSFMetric: YPerpPSFMetric
       plots.yPerpCModelPlot: YPerpCModelPlot
       metrics.yPerpCModelMetric: YPerpCModelMetric
+      plots.skyObjectHistPlot: SkyObjectHistPlot
```
laurenam, line 29: Is there a reason we are not also pushing metrics for the sky objects?  Oh, is that what `SkyFluxStatisticMetric` is?  If so, I’m not seeing where/how they are sharing code for consistent source selection, etc. (but that may just be because I’m still finding the structure of this repo to be quite mysterious…)

laurenam (approved):

pipelines/coaddQualityCore.yaml
leeskelvin: That's right, the `SkyFluxStatisticMetric` was put together by Jeff I believe, and currently the plotting and metric generation are not linked. They both do however make use of `SkyObjectSelector`, so selections _should_ be being generated consistently.

laurenam: Then do you think we _should_ make a ticket to get these synched up (my understanding is that the synching of plots & metrics was a primary driver for the creation & design of `analysis_tools`)?

leeskelvin: I don't have a high enough overview of `analysis_tools` to be able to make that call myself. I'm happy to defer to either yourself or @jeffcarlin for your views on this? For the purposes of this ticket however, I'm keen to focus on getting this into the regularly generated plotting pipeline as soon as possible.

laurenam: I too am keen to see the sky object plots make it into the core pipeline (and am almost certainly at a lower altitude than you on the `analysis_tools` overview scale 😉) , but I just worry about hastily pushing to `main` examples that do not follow the intended conventions (as we have seen in the past, these almost certainly will be copy-pasted/followed by others and then we end up with a mess).  I've approved, so feel free to merge, but perhaps it's worth also getting opinions from @yalsayyad / @natelust / @sr525 on how we should handle this going forward...

jeffcarlin: I agree that we should try to reuse/share code whenever possible. I think it'd be fine to merge this, and then have a second ticket to bring the plot and metric more in line with each other.

leeskelvin: Thanks @laurenam and @jeffcarlin. Following this thread, plus discussions we had yesterday, I've also now added sky source sky and hist plots on this ticket. Sky object sky plots have also been added.

PR #47, opened 2022-12-02
DM-36797: Add RhoStatistics

arunkannawadi:


python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+                colYy=self.colPsfYy,
+                colXy=self.colPsfXy,
+                sizeType=self.sizeType,
+            ),
+        )
+
+        # chi-type ellipticity has a shear response of 2, so we need to
+        # divide by 2 so that the rho-stats do not depend on the
+        # ellipticity-type.
+        responsitivity = 2.0 if self.ellipticityType == "chi" else 1.0
```
arunkannawadi, line 486: The introduction of `responsitivity` is new and was not present in `analysis_drp`. This ensures that the rho statistics always corresponds to the additive contamination to shear-shear correlation function regardless of the ellipticity type we choose.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -323,6 +324,16 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
         return cast(Vector, data[key])
 
 
+class PsfStarSelector(FlagSelector):
+    bands = ListField[str](
+        doc="The bands to apply the flags in, takes precedence if band supplied in kwargs",
+        default=["i"],
+    )
```
sr525, line 331: Is this what you want the default behaviour to be? If you want it to default to the band the plot is being made in then set this to [].

python/lsst/analysis/tools/tasks/objectTableSurveyAnalysis.py
```diff
@@ -0,0 +1,119 @@
...
+            tableName = inputs[connectionName][0].ref.datasetType.name
+            run = inputs[connectionName][0].ref.run
+
+        plotInfo = {"tableName": tableName, "run": run}
+
+        if dataId is not None:
+            for dataInfo in dataId:
+                plotInfo[dataInfo.name] = dataId[dataInfo.name]
+
+        return plotInfo
```
sr525, line 89: Does the plotInfo function in plotUtils not work for this? They look pretty much identical except that the standard one has more information in it.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+    are identical to the default values set in v4.2 of `treecorr`. The
+    parameters that are excluded in this class are
+    'verbose', 'log_file', 'output_dots', 'rng' and 'pairwise' (deprecated).
+    For details about these options, see the documentation for `treecorr`:
+    https://rmjarvis.github.io/TreeCorr/_build/html/correlation2.html
+
+    A separate config class is used instead
+    of constructing a `~lsst.pex.config.DictField` so that mixed types can be
+    supported and the config can be validated at the beginning of the
+    execution. The ``toDict`` provides a minimal dictionary that override only
```
sr525, line 64: overrides

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+    the default values and excludes the key-values pairs when the item is None.
```
sr525, line 65: values -> value

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+
+    max_sep = Field[float](
+        doc=(
+            "The maximum separation in units of sep_units, if relevant. "
+            "Exactly three of nbins, bin_size, min_sep, max_sep are required. "
+            "If max_sep is not given, it will be calculated from the values "
+            "of the other three."
+        ),
+        optional=True,
+    )
```
sr525, line 114: What happens if three of them aren't set?

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+        doc=(
+            "Whether to use the 'brute force' algorithm? "
+            "Unlike treecorr, setting this to 1 or 2 is not supported."
+        ),
+        default=False,
+        optional=True,
+    )
+
+    split_method = ChoiceField[str](
+        doc=("How to split the cells in the tree when building the tree " "structure."),
```
sr525, line 150: Odd formatting, tree " "structure."

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+    def validate(self):
+        # Docs inherited from base class
+        super().validate()
+        req_params = (self.nbins, self.bin_size, self.min_sep, self.max_sep)
+        num_req_params = sum(param is not None for param in req_params)
+        if num_req_params != 3:
+            msg = (
+                "You must specify exactly three of ``nbins``, ``bin_size``, ``min_sep`` and ``max_sep``"
+                f" in treecorr_config. {num_req_params} parameters were set instead."
+            )
```
sr525, line 329: I guess this answered my question, would it be useful though to have defaults set for the parameters?

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: Assuming it gets through to TreeCorr, then TreeCorr would raise an exception in that case.  Not sure if that's what you want here or not.

python/lsst/analysis/tools/actions/plot/xyPlot.py
```diff
@@ -0,0 +1,170 @@
...
+        allowed={scale: scale for scale in ("linear", "log", "symlog")},
+    )
+
+    yScale = ChoiceField[str](
+        doc="The scale to use for the y-axis.",
+        default="linear",
+        allowed={scale: scale for scale in ("linear", "log", "symlog")},
+    )
+
+    xLinthresh = Field[float](
```
sr525, line 80: Inconsistent camel case

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: Ah, `black` seems to have converted two-line docs into single line ones (rightly so) but without removing the extra quotations.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: I don't think a reasonable defaults can be set that is agnostic to the data/problem at hand. I believe that is why `treecorr` does not have defaults for these either. The default values for all other fields correspond to the default values of the corresponding quantities in `treecorr`. So for the sake of consistency, I'd like to leave them without any defaults.

python/lsst/analysis/tools/actions/plot/xyPlot.py
arunkannawadi: `xLinThresh`?

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: But it wouldn't get to `treecorr` because an exception would be raised here: https://github.com/lsst/analysis_tools/blob/041b14cb1ecba793d69d79cca0623ef01af8db96/python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py#L323-L330

This duplicates what `treecorr` does, but that is desirable because we want misconfiguration to be captured at config time, not at runtime, which is what would happen if we let `treecorr` handle it.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
sr525: This is present in a couple of places.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
sr525: I saw the catch below, makes sense now.

python/lsst/analysis/tools/actions/plot/xyPlot.py
sr525: For the others as well.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+    )
+
+    brute = Field[bool](
+        doc=(
+            "Whether to use the 'brute force' algorithm? "
+            "Unlike treecorr, setting this to 1 or 2 is not supported."
+        ),
+        default=False,
+        optional=True,
+    )
```
rmjarvis, line 147: Pretty sure you wouldn't ever want this for rho stats...

rmjarvis: In fact, a bunch of these are irrelevant for this use case, so should probably be removed here.  Everything from here down to num_bootstrap I don't think should ever be changed from their default values.  (Many of them aren't even relevant for rho stats, so wouldn't do anything if you did change them.  For the others, it's just a bad idea.)

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+
+    num_threads = Field[int](
+        doc=(
+            "How many OpenMP threads to use during the calculation. "
+            "If set to None, use all the available CPU cores."
+        ),
+        default=None,
+        optional=True,
+    )
```
rmjarvis, line 318: I'm not sure if you want num_threads to be settable either.  Probably that should be set based on something in the DM configuration elsewhere, rather than let the user set it explicitly here, but I don't know enough to say what to recommend on that one.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+
+    Notes
+    -----
+    The exact definitions of rho statistics as defined in [1]_ are given below.
+    In addition to these five, we also compute the auto-correlation function of
+    the fractional size residuals and call it as the :math:`\rho_0( \theta )`.
+
+    .. math::
+
+       \rho_0(\theta) &= \left\langle \frac{\delta T_{PSF}}{T_{PSF}}(x) \frac{\delta T_{PSF}}{T_{PSF}}(x+\theta) \right\rangle  # noqa: E501, W505
```
rmjarvis, line 365: rho0 is traditionally what people call <e*_psf e_psf>.  So I think using this name for a different statistic is likely to be confusing.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+       \rho_2(\theta) &= \langle e^*_{PSF}(x) \delta e_{PSF}(x+\theta) \rangle
+
+       \rho_3(\theta) &= \left\langle (e^*_{PSF}\frac{\delta T_{PSF}}{T_{PSF}}(x)) \delta e_{PSF}(x+\theta) \right\rangle  # noqa: E501, W505
+
+       \rho_4(\theta) &= \left\langle (\delta e^*_{PSF}(x) (e_{PSF}\frac{\delta T_{PSF}}{T_{PSF}}(x+\theta)) \right\rangle  # noqa: E501, W505
+
+       \rho_5(\theta) &= \left\langle (e^*_{PSF}(x) (e_{PSF}\frac{\delta T_{PSF}}{T_{PSF}}(x+\theta)) \right\rangle  # noqa: E501, W505
+
+    The definition of ellipticity used in [1]_ correspond to ``epsilon``-type ellipticity, which is typically
+    smaller by a factor of 4 than using ``chi``-type ellipticity.
```
rmjarvis, line 378: epsilon and chi are not standard names for what you're referring to here.  The more typical names are shear vs distortion.  epsilon is sometimes used to refer to shear, and chi is sometimes distortion.  But not by any stretch universally.  So you should at least give the names in addition to (or instead of) the Greek letters.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+
+    colPsfXy = Field[str](
+        doc="The column name to get the PSF xy shape component from.", default="{band}_ixyPSF"
+    )
+
+    ellipticityType = ChoiceField[str](
+        doc="The type of ellipticity to calculate",
+        allowed={
+            "chi": "Distortion, defined as (Ixx - Iyy)/(Ixx + Iyy)",
+            "epsilon": ("Shear, defined as (Ixx - Iyy)/" "(Ixx + Iyy + 2*sqrt(Ixx*Iyy - Ixy**2))"),
```
rmjarvis, line 415: That's not actually the "definition" of shear.  That's just one way to estimate shear.  Shear is defined by the elements in the 2x2 transformation matrix.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: This is only right in the limit that e is small.  That's probably ok for this purpose, but the distortion responsivity is actually 2-<|e|^2>.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: Also, as per comment above, I'd prefer the "type" names be distortion and shear, rather than semi-arbitrary Greek letters.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -0,0 +1,649 @@
...
+            # Calculate the auto-correlation
+            xy.process(catA)
+        else:
+            catB = treecorr.Catalog(
+                ra=ra, dec=dec, g1=g1b, g2=g2b, ra_units=raUnits, dec_units=decUnits, logger=_LOG
+            )
+            # Calculate the cross-correlation
+            xy.process(catA, catB)
+
+        _LOG.debug("Correlated %d pairs out of %d possible pairs", xy.npairs, len(ra) * (len(ra) - 1) / 2)
```
rmjarvis, line 648: I would expect this to give an error, although maybe it's just a logging warning.  `xy.npairs` is an array, not an int.  It's the number of pairs for each bin of the output correlation.  It's also a bit weird to report this as being how many were correlated out of some total possible.  The ones that weren't correlated were at different separations, so were they "possible"?

python/lsst/analysis/tools/analysisParts/rhoStatistics.py
```diff
@@ -0,0 +1,77 @@
...
+            vectorKey="patchWhole", selector=VectorSelector(vectorKey="starSelector")
+        )
+
+        self.process.calculateActions.stars = CalcRhoStatistics()
+
+        self.process.calculateActions.stars.treecorr.nbins = 10
+        self.process.calculateActions.stars.treecorr.min_sep = 0.1
+        self.process.calculateActions.stars.treecorr.max_sep = 100.0
+        self.process.calculateActions.stars.treecorr.sep_units = "arcmin"
+        self.process.calculateActions.stars.treecorr.metric = "Arc"
```
rmjarvis, line 77: Don't use Arc.  It's just slower and doesn't add anything useful compared to the default Euclidean.

rmjarvis (changes requested):

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: The `e` there is intrinsic ellipticity, isn't it? The point source is essentially round, so its intrinsic ellipticity is zero. Any way, this is also set to match what was used with the `reGauss` shapes in the HSC weak lensing analysis code.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: rho6 then?

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: Ah, good catch. I meant to put in `sum(xy.npairs)` there.

arunkannawadi: The total number here is to give some indication about whether we have explore sufficient range in the scales or not. As we plan to run this on a regular cadence as we collect data, the angular scales that we set might evolve and this is intended to guide that.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: I agree that `num_threads` should probably be removed, but this class is meant to map all the keywords that can be passed to instantiate any `BinnedCorr2` instances more generally. I don't have a use case beyond rho statistics for now, so I could go either way.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: No, it's the PSF ellipticity.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: I don't think the total of npairs, either as an absolute number or as compared to N*(N-1)/2, would say much about how well you're covering scales though.  Maybe S/N of the first and last bin?  If the S/N of either gets largish, it could mean that it would be useful to extend the range.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -323,6 +324,16 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+
+    def setDefaults(self):
+        self.selectWhenTrue = ["{band}_calib_psf_used"]
+
```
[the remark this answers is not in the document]
sr525: It would replace the entire psf selector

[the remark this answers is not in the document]
sr525: It would replace the entire psf selector

python/lsst/analysis/tools/tasks/objectTableSurveyAnalysis.py
sr525: I would have thought that this would come up more regularly, it might be good to generalise the plotUtils one.

arunkannawadi: So, generalize the `parsePlotInfo` in utils, remove the method from `tasks/base.py` and use the utility function both here and in `tasks/base.py`? That sounds like a separate ticket to me.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: Is there any publication where rho0 is defined?

python/lsst/analysis/tools/tasks/objectTableSurveyAnalysis.py
arunkannawadi: DM-37419

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: I don't know if I've seen it in anything published.  But the SLAC people who do PSF work use that notation pretty regularly.

rmjarvis: Marco used it here: https://arxiv.org/pdf/2011.03408.pdf  Figure 8.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: For HSC validation, the `responsitivity` has been set to 2 exactly. Given that the next big validation run will be on HSC, I'd like to keep it this way for consistency.  We can make it more accurate later on, but for LSST, we'd likely be using shear estimator, and not distortion, so it shouldn't matter much any way.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: Removed a bunch of fields incl. `num_threads`, because it is easier to add fields and even backport them, than to remove fields.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
rmjarvis: So, you want to validate that you get the same (slightly) wrong answer...  OK, I guess.  Wouldn't be my choice though.  I'd calculate the ratio you expect to be different because of this as part of the validation, with comments saying why the HSC values are a little off.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
arunkannawadi: As a quantity not involving any PSF residuals, I'm unhappy with the use of `rho0` to mean ellipticity correlations. I could go with `rho6` although I suspect that'd prove to be a bad choice as well. I could name it something like `mu0` but that go lead this plot becoming undiscoverable :-|

rmjarvis: Maybe just call it dT_auto or something more direct like that, rather than some code letter/number?  Also, what's the intended use case for it?  The mean <dT> (one-point function) is relevant for WL systematics, but I don't think this combination is directly applicable to WL.  Does LSS care about this?

arunkannawadi: The auto-correlation function of dT/T is what was called as `rho3` in [Melchior et al. (2015)](https://ui.adsabs.harvard.edu/abs/2015MNRAS.449.2219M/abstract), which you denoted as `rho'3` statistic in your 2016 paper. The reason why it is interesting here, not necessarily from a scientific standpoint, but from a pipelines standpoint is that, for reasons not clear to us, we see a significant offset in PSF size on coadds. This serves as a good diagnostic tool to see how our planned changes to PSF coaddition improves the situation. This complements our other tools involving scatter plots and other one-point calculations from it.

Perhaps, I'll just follow your notation call it `rho3prime` or something.

rmjarvis: Ah, yes.  Peter and I disagreed at first about how dT errors propagate to xi+.  Originally he thought his `rho3 * <|e|^2>` was the right systematic error on xi+.  But I convinced him that the e terms couldn't be brought out of the angle brackets.  They needed to stay as what I ended up calling rho3.  `<e dT/T, e dT/T>`.

python/lsst/analysis/tools/tasks/objectTableSurveyAnalysis.py
sr525: I don't think not duplicating code unnesecarily is really another ticket. Having two of these will mean that some plots might end up displaying different things and the plot info generation will get out of sync.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
sr525: Just adding a vote in for understandable names not random greek letters. Remember that non WL people will see this, like me, and it would be great if it was understandable.

python/lsst/analysis/tools/tasks/objectTableSurveyAnalysis.py
arunkannawadi: There is already code duplication in the `main` branch. The ticket is to deduplicate that. And this `plotInfo` will be a bit different for this task because of multiple tracts in it, and even more so when [DM-37330](https://jira.lsstcorp.org/browse/DM-37330?) lands. It is okay in the object-oriented paradigm for simple and short methods like these to be overridden for special cases, without worrying about partial code duplication. I strongly prefer that over the base method handling all possible cases. Any additions to the base method, that applies to this task as well, has to be added here as well. If we miss that, we will catch it in one of the metrics meetings and can quickly add it here as well.

arunkannawadi (commented):
OK, I believe I have addressed all the comments, or created TODO ticket items to work on. This is certainly a minimum viable thing that I'd like to get this in the weekly that gets tagged tonight. @rmjarvis do I have your approval?

rmjarvis (approved):

rmjarvis:
> @rmjarvis do I have your approval?

Sure.  Looks fine to me.

PR #50, opened 2022-12-15
DM-37158: Add 'Plot' to current plot type names in a-tools

leeskelvin:


python/lsst/analysis/tools/tasks/reconstructor.py
```diff
@@ -20,9 +20,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
```
natelust, line 22: @leeskelvin can you add an `__all__`

python/lsst/analysis/tools/tasks/reconstructor.py
```diff
@@ -95,3 +97,46 @@ def reconstructAnalysisTools(
...
+        who's plots are to be queried. If no label is given, the entire
+        collection will be queried.
+
+    Yields
+    ------
+    plotName : `str`
+        Dataset type name of the plot.
+    """
+    if isinstance(collections, str):
+        collections = [collections]
```
parejkoj, line 127: You don't need this here: `butler.get` and `queryDatasets` both handle it for you.

python/lsst/analysis/tools/tasks/reconstructor.py
```diff
@@ -95,3 +97,46 @@ def reconstructAnalysisTools(
...
+    Parameters
+    ----------
+    butler : `~lsst.daf.butler.Butler`
+        The butler where the data is stored.
+    collections : `str` or `list` [`str`]
+        Collection(s) within the butler associated with the desired data.
+    label : `str`, optional
+        The label from the `~lsst.pipe.base.Pipeline` associated with the task
+        who's plots are to be queried. If no label is given, the entire
+        collection will be queried.
```
parejkoj, line 119: "the entire collection" -> "all requested collections"

python/lsst/analysis/tools/tasks/reconstructor.py
```diff
@@ -95,3 +97,46 @@ def reconstructAnalysisTools(
...
+    Yields
+    ------
+    plotName : `str`
+        Dataset type name of the plot.
+    """
+    if isinstance(collections, str):
+        collections = [collections]
+    if label is not None:
+        configDSType = f"{label}_config"
+        configs = [butler.get(configDSType, collections=collections)]
```
parejkoj, line 130: I think I'd just remove the "configDSType" variable from here, since it's fairly trivial.

parejkoj (changes requested):
So it looks like we're not enforcing that plots are named "Plot", just changing existing ones to be named such? Where would we document the convention that analysis_tools plots all have names ending in "Plot", to help others making new datasets?

Please add some tests that demonstrate that this behaves as desired; I think I understand the logic, but a test that makes a demo butler and puts some things in it to check that this only gets the plots and nothing else would be good.

leeskelvin:
Thanks @parejkoj. Requested changes merged into this PR. Please see Jira for more discussion on documentation.

tests/test_getPlotDatasetTypeNames.py
```diff
@@ -0,0 +1,80 @@
...
+    def testGetPlotNames(self) -> None:
+        """Test that we can get plot dataset type names from a butler."""
+        plotConfig = ObjectTableTractAnalysisConfig()
+        plotConfig.plots.skyObjectHistPlot = SkyObjectHistPlot()
+        otherConfig = Config()
+        _ = self.butler.put(plotConfig, "plotType_config")
+        _ = self.butler.put(otherConfig, "otherType_config")
+        plots = getPlotDatasetTypeNames(self.butler, "test_run")
+        for plot in plots:
+            self.assertIn("skyObjectHistPlot", plot)
```
parejkoj, line 67: I'm confused by this loop: `getPlotDatasetTypeNames` should return a list with exactly one item, so you should be able to do `self.assertEqual(["skyObjectHistPlot", plots], with no loop, no?

tests/test_getPlotDatasetTypeNames.py
```diff
@@ -0,0 +1,80 @@
...
+        self.butler.registry.registerDatasetType(self.datasetTypePlot)
+        self.datasetTypeOther = DatasetType(
+            "otherType_config", dimensions=(), storageClass="Config", universe=self.butler.registry.dimensions
+        )
+        self.butler.registry.registerDatasetType(self.datasetTypeOther)
+
+    def tearDown(self):
+        removeTestTempDir(self.root)
+
+    def testGetPlotNames(self) -> None:
```
parejkoj, line 58: This is a good test, but you also should have a test of the `label` if-branch of the code.

parejkoj, line 58: I'm not sure anything is gained by adding type annotations to unittests: they all return `None` by definition.

tests/test_getPlotDatasetTypeNames.py
```diff
@@ -0,0 +1,80 @@
...
+        """Test that we can get plot dataset type names from a butler."""
+        plotConfig = ObjectTableTractAnalysisConfig()
+        plotConfig.plots.skyObjectHistPlot = SkyObjectHistPlot()
+        otherConfig = Config()
```
parejkoj, line 62: You should have another Config that is an `AnalysisBaseConfig` but not a `Plot`, since there's a branch for that in the code.

parejkoj (commented):
I recommend checking the coverage output for your new code: I see a few branches that aren't covered by your tests here (I think I commented on all of them, but the pytest-cov output will tell you).

If you're not familiar with the coverage output, I wrote it up in the docs here:

https://developer.lsst.io/stack/unit-test-coverage.html#python

tests/test_getPlotDatasetTypeNames.py
leeskelvin: Try running it yourself and you will see, by default, that this list contains 6 elements: one for each band.

parejkoj: Oh, hmm... In that case I'd probably make that the explicit list I'm testing against with `assertEquals`. You know exactly what the output should be, just test it directly.

leeskelvin:
Thanks again @parejkoj. Running `pytest -sv --cov-report html:htmlcov-gen2 --cov python tests/test_getPlotDatasetTypeNames.py` reports checkers green. All other suggestions merged, and PR updated.

tests/test_getPlotDatasetTypeNames.py
leeskelvin: What benefit does this have relative to its current set up? I can't see it myself, and as this unit test is already relatively complex, I'm inclined to leave this as-is.

parejkoj: When you know the *exact* output that a function should have, you should test that exact output. Currently, there are a variety of things that would pass this test, but aren't actually what you're expecting (for example, `["g_skyObjectHistPlot", "g_skyObjectHistPlot"]` would pass, but I'm pretty sure is incorrect). It actually simplifies the test code, too: there's no loop with `assertIn` just an explicit "here's what the answer should be".

natelust: In this case @leeskelvin is right to leave it as an `assertIn`, as the exact bands depend on what the default configs of the task are, which is outside the unit test. One could go though and rigorously construct everything such that all inputs are known exactly, but it adds nothing to the concreteness of the test for the behavior in the function, and increases the complexity, and future maintenance of the test.

tests/test_getPlotDatasetTypeNames.py
```diff
@@ -0,0 +1,113 @@
...
+        """Test that we can get plot dataset type names prefixed with a certain
+        label from a given collection in a butler."""
+        plotConfig = ObjectTableTractAnalysisConfig()
+        plotConfig.plots.skyObjectHistPlot = SkyObjectHistPlot()
+        baseConfig = AnalysisBaseConfig()
+        baseConfig.connections.outputName = "baseName"
+        otherConfig = Config()
+        _ = self.butler.put(plotConfig, "plotLabel_config")
+        _ = self.butler.put(baseConfig, "baseLabel_config")
+        _ = self.butler.put(otherConfig, "otherLabel_config")
```
parejkoj, line 97: I think these other two should also have plot config types, so that you're clearly checking the "did I only get plots with name `plotLabel`" logic.

parejkoj (approved):
One more comment about the new test.

PR #51, opened 2022-12-15
DM-34625: Add sky coverage plots of visit summary quantities to analysis_tools

laurenam:


cmsaunders (approved):
A few overall comments:

- Can you add the units for the colorbars?
- From the naming, i.e. `VisitCoveragePlot`, it isn't really clear that the plots are for multiple visits, and I think users might get mixed up with the visit-level plots. Maybe something like `MultiVisitCoveragePlot`?
- Can you explain the "focalPlane" vs "raDec" plots options somewhere? Without looking at the examples, it isn't obvious that the focalPlane plot will give one number per detector, while the raDec is binned in ra/Dec.

I didn't go through the big `makePlot` function as closely as I would have liked to, but I wanted to get this back to you before I go on break. I would be happy to look at it again, but if you want to get this merged, I will let you make the call on that.

laurenam (commented):
> Can you add the units for the colorbars?

Added (where the parameters are not unitless).

> From the naming, i.e. VisitCoveragePlot, it isn't really clear that the plots are for multiple visits, and I think users might get mixed up with the visit-level plots. Maybe something like MultiVisitCoveragePlot?

Suggestion adopted and action/plot filename changed to `multiVisitCoveragePlot.py`.

> Can you explain the "focalPlane" vs "raDec" plots options somewhere? Without looking at the examples, it isn't obvious that the focalPlane plot will give one number per detector, while the raDec is binned in ra/Dec.

They both plot the same data, just in different "projections", which I hoped would be clear with a config name of `projection`.  To (hopefully) help, I added the line:
"In either case, one point is plotted per visit/detector combination." 
to the docstring.

arunkannawadi:
Coming in here super late, but I don't see `CcdVisitTableAnalysisTask` called anywhere (neither in this PR not in the `main` branch now). I see that the `coveragePlots.yaml` that Lauren attached to the Jira ticket uses it, but I don't see it in the `pipelines/` folder here. Does that mean we no longer use this, and that it might be bitrotting?

laurenam:
Sigh...indeed, you are correct (and thank you very much for pointing this out!)  I definitely still want the ability to make these plots, so I guess I should make a ticket to (fix any bitrot) and add a pipeline to the repo.  Do you know how tasks that aren't necessarily run in the "core" pipelines get tested/CI'd?

arunkannawadi:
Oh, I definitely want to reuse this task but found that I had to comment out one line to be able to run it (hence the bitrot comment). I think having a pipeline that could be run would be a good start. If you want it CI'd, you probably need a test with made up data similar to this one: https://github.com/lsst/analysis_tools/blob/main/tests/test_catalogMatch.py

laurenam:
Ah, ok, thanks!  How would you like to proceed?  Shall I make a ticket as I suggest above, or will your pipeline & bitrot fix(?) be suitable (modulo the ci option...)?

arunkannawadi:
I think you should add the `coveragePlots.yaml` file to `pipelines/` regardless (unless you have strong reasons not to?). That's likely to happen faster than my new ticket (DM-41134) but I am okay with whatever happens first.

laurenam:
Sounds good.  I'll try to get to that soon...ish!

PR #58, opened 2023-01-11
DM-37330: Add a utility function to compress tract list

arunkannawadi:
This PR adds a utility function to `plotUtils` and unit tests for it. It also copies the unit tests for other `plotUtils` functions from `analysis_drp`.

### DIFF
--- a/python/lsst/analysis/tools/actions/plot/plotUtils.py
+++ b/python/lsst/analysis/tools/actions/plot/plotUtils.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 __all__ = ("PanelConfig",)
 
-from typing import TYPE_CHECKING, List, Mapping, Tuple
+from typing import TYPE_CHECKING, Iterable, List, Mapping, Tuple
 
 import matplotlib
 import matplotlib.pyplot as plt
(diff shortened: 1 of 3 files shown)


tests/test_plotUtils.py
```diff
@@ -0,0 +1,97 @@
...
+
+import unittest
+
+import lsst.utils.tests
+import numpy as np
+from lsst.analysis.tools.actions.plot.plotUtils import perpDistance, shorten_list, stellarLocusFit
+
+
+class FitTest(lsst.utils.tests.TestCase):
+    """Test to see if the fitting and distance calculations are working"""
```
leeskelvin, line 29: Missing period.

tests/test_plotUtils.py
```diff
@@ -0,0 +1,97 @@
...
+        # line. Test that these are returning what is expected.
+        self.assertFloatsAlmostEqual(paramsOut["mODR"], 1.0)
+        self.assertFloatsAlmostEqual(paramsOut["mODR2"], 1.0)
+        self.assertFloatsAlmostEqual(paramsOut["bODR"], 0.0)
+        self.assertFloatsAlmostEqual(paramsOut["bODR2"], 0.0)
+        self.assertFloatsAlmostEqual(paramsOut["bPerpMin"], 0.0)
+        self.assertFloatsAlmostEqual(paramsOut["bPerpMax"], 22.0)
+
+    def testPerpDistance(self):
+        """Test the calculation of the perpendicular distance"""
```
leeskelvin, line 54: Missing period.

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -598,6 +598,68 @@ def addSummaryPlot(fig, loc, sumStats, label):
     return fig
 
 
+def shorten_list(numbers: Iterable[int]) -> str:
+    """Shorten a list of integers.
+
+    Parameters
+    ----------
+    numbers : `list` [`int`]
```
leeskelvin, line 606: Rather than ``list``, should this just be `iterable` (without back ticks)? From the doc string and the unit test, it looks like you're happy accepting any iterable?

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -598,6 +598,68 @@ def addSummaryPlot(fig, loc, sumStats, label):
...
+    >>> shorten_list([1,2,3,5,6,8])
+    "1-3,5-6,8"
+
+    >>> shorten_list((1,2,3,5,6,8,9,10,11))
+    "1-3,5-6,8-11"
+
+    >>> shorten_list(range(4))
+    "0-3"
+    """
+    if not numbers:  # empty container
```
leeskelvin, line 625: As we discussed:
```
if len(numbers) == 0:
```
will catch numpy array inputs.

leeskelvin: Alternatively, perhaps convert to a `set` right at the top of this function. This would then allow for the code block checking for possible duplicate number entires in your input iterable to be dropped.

leeskelvin: Option #3: you could swap the order of `len(numbers)` and `sorted` such that sorting occurs first. This will guarantee that your numbers iterable is in a list format, which would then have a `len`. Otherwise, the input (e.g., an iterator) may not have a `len` to pass the initial `if` check.

leeskelvin (approved):

python/lsst/analysis/tools/actions/plot/plotUtils.py
arunkannawadi: I think I am going to go with option 3. If it is an empty container, calling `sorted` on it is a trivial operation. OTOH, creating a set out of a large container might be relatively more expensive. Option 1 is still the best for our use case, but I am going to refrain from assuming that __len__ is implemented for ``numbers``.

arunkannawadi:
I went with option 3, for reasons I indicated in the inline comments. I also added two keyword-only arguments that can help customize this utility function. I have updated the examples and tests to illustrate that as well.

PR #60, opened 2023-01-17
DM-36444: photRepeatHistPlot and Configurable stats panel

psferguson:
The main change in this ticket is adding the ability to specify the stats to be shown on hist plot. They can be any scalar that has a vector key in the data (calculated in the body of the metric/plot). If no stats are specified this behaves as before where the N, median, and sigma mad are shown.


python/lsst/analysis/tools/actions/plot/histPlot.py
```diff
@@ -40,6 +49,29 @@
 log = logging.getLogger(__name__)
 
 
+class HistStatsPanel(Config):
```
sr525, line 52: Class doc string please, pretty please. I'm bored of adding them everywhere.

PR #61, opened 2023-01-18
DM-35207-v25: backport UpdateVisitSummaryTask and usage of its finalVisitSummary output

TallJimbo:


TallJimbo:
I'm ignoring the mypy lint for the backport here because I recognize it as being due to the action still targeting `main` or some other recent (non-v25) weekly of pipe_base.

PR #62, opened 2023-01-21
DM-37635: Cleanup analysis_tools to fix pipelines docs build failure

parejkoj:


python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -20,6 +20,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
+__all__ = ("AssociatedSourcesTractAnalysisTask", "AssociatedSourcesTractAnalysisConfig")
```
ktlim, line 23: Maybe the other order to match the code?

python/lsst/analysis/tools/tasks/ccdVisitTableAnalysis.py
```diff
@@ -20,7 +20,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-__all__ = ("CcdVisitTableAnalysisTask",)
+__all__ = ("CcdVisitTableAnalysisTask", "CcdVisitTableAnalysisConfig")
```
ktlim, line 23: Order again here.

python/lsst/analysis/tools/tasks/base.py
```diff
@@ -31,6 +31,8 @@
 connection classes and should specify a unique name
 """
 
+__all__ = ("AnalysisBaseConfig", "AnalysisPipelineTask")
```
ktlim, line 34: Is `AnalysisBaseConnections` needed as well?  See diaObjectTableAssocAnalysis.py and others.

ktlim (approved):

python/lsst/analysis/tools/tasks/base.py
parejkoj: I've generally assumed that Connections classes should not be part of the public API, since you never need to import them as far as I'm aware. Configs do get imported and modified. I've asked on #dm-science-pipelines about this.

PR #63, opened 2023-01-25
DM-37635: Skip type definitions that break the docs build

parejkoj:


jonathansick (approved):
Looks good. I'm not entirely sure what the issue with these might be; possibly it can be fixed when we upgrade the documentation stack for pipelines.lsst.io

PR #67, opened 2023-02-08
DM-37927: Make calibrateConfig argument optional

yalsayyad:


python/lsst/analysis/tools/actions/plot/multiVisitCoveragePlot.py
```diff
@@ -301,7 +301,10 @@ def makePlot(
         if len(notFoundKeys) > 0:
             raise RuntimeError(f"Requested column(s) {notFoundKeys} is(are) not in the data table.")
 
-        maxMeanDistanceArcsec = calibrateConfig.astrometry.maxMeanDistanceArcsec  # type: ignore
+        if calibrateConfig is None:
+            maxMeanDistanceArcsec = 0.0
```
laurenam, line 305: I think setting this to 0.0 will cause problems here:
https://github.com/lsst/analysis_tools/pull/67/files#diff-0c5832757a401b77baae92ff1a4b428728fba6c3c14fec79fbac1230d293695fL424

PR #74, opened 2023-03-20
DM-38372: Switch to numpy random generator for test_catalogMatch and add seed

cmsaunders:


tests/test_catalogMatch.py
```diff
@@ -32,6 +32,8 @@
 from lsst.daf.butler import DatasetRef, DatasetType, DimensionUniverse, StorageClass
 from lsst.meas.algorithms import ReferenceObjectLoader
 
+rng = np.random.default_rng(12345)
```
natelust, line 35: Since you are no longer relying on modifying global numpy random behavior and are using an instance of a random generator, can you move this to be an instance variable of the class you use it in?

natelust (approved):
See my one comment. Also since you switched to the new style of an instance generator, can you squash  away the commit setting the seed?

cmsaunders:
Thanks! Yes, I'll squash. I should have said that I was planning to squash once the review was done.

PR #75, opened 2023-04-05
DM-37918: Structural changes to the package

natelust:
This is just a draft so people can view the changes easily, and comment on it. It is not intended to be taken as the final product (or even fully working)


bechtol (approved):
Changes all look good. I've added several more design-oriented comments on the ticket that I think could be addressed on future tickets.

pipelines/coaddQualityCore.yaml
```diff
@@ -5,41 +5,30 @@ tasks:
...
-        from lsst.analysis.tools.analysisPlots import *
-        from lsst.analysis.tools.analysisMetrics import *
+        from lsst.analysis.tools.atools import *
         from lsst.analysis.tools.contexts import *
   analyzeObjectTableSurveyCore:
     class: lsst.analysis.tools.tasks.ObjectTableSurveyAnalysisTask
     config:
       connections.outputName: objectTableSurvey
-      plots.rhoStatisticsPlot: RhoStatisticsPlot
+      plots.rhoStatistics: RhoStatistics
```
arunkannawadi, line 28: Should this have changed to `atools` from `plots`?

natelust: Yeah, missed one. I left an alias in place so it would not be broken now anyway, but it is best to change it

arunkannawadi: Okay, just wanted to confirm. I'll change it as part of DM-38770.

PR #80, opened 2023-04-18
DM-37703: Use InMemoryDatasetHandle and drop DatasetRef from test

timj:


tests/test_catalogMatch.py
```diff
@@ -31,32 +31,7 @@
...
-        Parameters
-        ----------
-        parameters : `dict`, optional
-            Parameter dictionary.  Supported key is ``columns``.
-        Returns
-        -------
-        dataframe : `pandas.DataFrame`
-            dataframe, cut to the specified columns.
-        """
-        if "columns" in parameters:
```
timj, line 55: It doesn't look like this parameter is ever used because it looks like this code has been copied from an implementation that had been used for pandas DataFrame objects -- the tests pass and SimpleCatalog is not configured to understand a columns parameter.

tests/test_catalogMatch.py
```diff
@@ -65,10 +39,6 @@ class MockDataId:
     def __init__(self, region):
         self.region = region
 
-        datasetDimensions = DimensionUniverse().extract(["htm7"])
-        datasetType = DatasetType("gaia_dr2_20200414", datasetDimensions, StorageClass("SimpleCatalog"))
-        self.ref = DatasetRef(datasetType, {"htm7": "mockRefCat"})
```
timj, line 70: Turns out this ref wasn't needed at all so I simply deleted it.

PR #85, opened 2023-05-03
DM-39004: Add simple doc strings for all classes/variables.

leeskelvin:


python/lsst/analysis/tools/actions/plot/rhoStatisticsPlot.py
```diff
@@ -39,6 +39,8 @@
 
 
 class RhoStatisticsPlotAction(PlotAction):
+    """A class to plot rho statistics."""
+
```
sr525, line 43: Arun has just fixed a few of these, just make sure you don't overwrite them on rebase.

sr525 (approved):

PR #87, opened 2023-05-03
DM-39005: Restructure analysis tools docs landing page.

leeskelvin:


doc/lsst.analysis.tools/index.rst
```diff
@@ -29,53 +26,32 @@ For a tutorial on working with
...
-.. _lsst.analysis.tools-contributing:
-
-Contributing
-============
-
-``lsst.analysis.tools`` is developed at https://github.com/lsst/analysis_tools.
-You can find Jira issues for this module under the `analysis_tools <https://jira.lsstcorp.org/issues/?jql=project%20%3D%20DM%20AND%20component%20%3D%20analysis_tools>`_ component.
-
-.. If there are topics related to developing this module (rather than using it), link to this from a toctree placed here.
+If you have any questions regarding ``analysis_tools`` it is recommended that you post your question on `The Community Forum <https://community.lsst.org/>`_.
```
cmsaunders, line 34: I just wanted to check that this was agreed on. Is someone assigned to look out for questions on Community, or do we expect that someone outside DRP is going to forward questions to us?

python/lsst/analysis/tools/__init__.py
```diff
@@ -21,19 +21,20 @@
 r"""This packages contains various analysis tools and PipelineTasks which
 run them.
 
-There are two types of primary tools within this package; `AnalysisMetrics
-and `AnalysisPlots`, both of which are specialized subclasses of `AnalysisTool`.
+There are two primary types of tools within this package: `AnalysisMetrics` and
+`AnalysisPlots`, both of which are specialized subclasses of `AnalysisTool`.
```
cmsaunders, line 25: There is `AnalysisTool` now, so this should be rewritten.

python/lsst/analysis/tools/__init__.py
```diff
@@ -21,19 +21,20 @@
...
 
-`AnalysisTool`\ s are objects that have three stages; prep, process, and
-post-process. Creating a new `AnalysisTool` is the process of choosing (though
-configuration) what `AnalysisAction` will run for each of those stages.
+An `AnalysisTool` object has three stages: prep, process, and produce.
+Creating a new `AnalysisTool` involves choosing (through
+configuration) an `AnalysisAction` to run for each of those stages.
```
cmsaunders, line 29: Since you can use multiple actions for each stage, how about "which `AnalysisAction`s to run..."

cmsaunders (approved):
This looks good. My biggest comment is to change the reference to `AnalysisPlot` and `AnalysisMetric`.

A couple of other things:
- Why do some of the plot types still not have hyperlinks?
- Were the docs for `nansigmaMad` and `sigmaMad` supposed to disappear?

doc/lsst.analysis.tools/index.rst
leeskelvin: My understanding is that the official line is to always redirect questions from the community onto Community. Slack is convenient, but also seen as a transient resource (i.e., not good for long-term information storage). Happy to punt on this one however if you would like to bring it up for a broader discussion?

python/lsst/analysis/tools/__init__.py
leeskelvin: Changed to (and merged with the next paragraph):
> The primary tool in this package is the `AnalysisTool` class.
> An `AnalysisTool` object has three stages: prep, process, and produce.
> Creating a new `AnalysisTool` involves choosing (through
> configuration) an `AnalysisAction` to run for each of those stages.

python/lsst/analysis/tools/__init__.py
leeskelvin: Modified the final sentence to read:
> Creating a new `AnalysisTool` involves choosing (through configuration) one or
> more `AnalysisAction` classes to run for each of those stages.

doc/lsst.analysis.tools/index.rst
leeskelvin: Update from Yusra via Slack:
> We can only direct staff to Slack. Your commissioning in-kinds are considered staff, so if you’re talking about comms to them in particular, you can direct them to Slack.  But if this is going on the [pipelines.lsst.io](http://pipelines.lsst.io/), you need to only direct them to community. That’s our only official user support.

leeskelvin: Removed the Slack component, and leaving Community here in this section accordingly.

leeskelvin:
Replies:

> Why do some of the plot types still not have hyperlinks?
I think this was a transient thing in the testing docs I built. I played around a few times, regenerating them, and they seem to work fine now. I'll keep an eye on this in the built docs to make sure all is well.

> Were the docs for nansigmaMad and sigmaMad supposed to disappear?
No, they were not, thank you for catching that. They've been added back in by adding the :include-all-objects: option under the primary automodapi directive :+1:

PR #93, opened 2023-05-05
DM-36968: Make per-tract survey property map plots

enourbakhsh:


python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+
+            # Set labels and legend.
+            ax2.set_xlabel(label)
+            ax2.set_ylabel("Normalized Count")
+            legend = ax2.legend(loc="best", frameon=False, fontsize=15)
+            for line, text in zip(legend.legendHandles, legend.get_texts()):
+                text.set_color(line.get_facecolor())
+
+            # Add nside info to `plotInfo`.
+            plotInfo["nside_coverage"] = mapData.nside_coverage
```
erykoff, line 309: The nside_coverage is an implementation detail of the map, and shouldn't be plotted.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+            plotInfo["nside_sparse"] = mapData.nside_sparse
```
erykoff, line 310: This should just be `nside` on the plotInfo I think.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+        -------
+        Results : `tuple` [`float`]
+            New extent as (new_lon_min, new_lon_max, new_lat_min, new_lat_max).
+        """
+        if n is None:
+            return None
+        lon_min, lon_max, lat_min, lat_max = fullExtent
+        lon_center, lat_center = (lon_min + lon_max) / 2, (lat_min + lat_max) / 2
+        half_lon = (lon_max - lon_min) / (2 * n)
+        half_lat = (lat_max - lat_min) / (2 * n)
```
erykoff, line 371: When doing the zoom, you want to make sure you scale the longitude (ra) range by cos(lat) so it works at a wide range of latitudes/declinations.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+
+    @staticmethod
+    def addTextToColorbar(cb, text, orientation="vertical", color="black", fontsize=14, fontweight="bold"):
+        """Helper method to add text inside the horizontal colorbar.
+
+        Parameters
+        ----------
+        cb : `matplotlib.colorbar.Colorbar`
+            The colorbar object.
+
```
erykoff, line 388: We don't need extra blank lines between parameter docstrings.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+        plotInfo: Optional[Mapping[str, str]] = None,
+        **kwargs,
+    ) -> Mapping[str, Figure]:
+        """Make the survey property map plot.
+
+        Parameters
+        ----------
+        data : `KeyedData`
+            The HealSparseMap to plot the points from.
+
```
erykoff, line 143: We don't need extra blank lines between docstring parameters.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+    @staticmethod
+    def getZoomedExtent(fullExtent, n):
+        """Get zoomed extent centered on the original full plot.
+
+        Parameters
+        ----------
+        fullExtent : `tuple` of `float`
+            The full extent defined by (lon_min, lon_max, lat_min, lat_max):
+
```
erykoff, line 344: We don't need the blank lines here.

python/lsst/analysis/tools/tasks/propertyMapTractAnalysis.py
```diff
@@ -0,0 +1,179 @@
...
+                setattr(
+                    self,
+                    f"{coaddName}Coadd_{propertyName}_{operationName}",
+                    ct.Input(
+                        doc=f"{operationLongName}-value map of {{propertyLongName}}",
+                        name=name,
+                        storageClass="HealSparseMap",
+                        dimensions=("tract", "skymap", "band"),
+                        multiple=False,
+                        deferLoad=False,
```
erykoff, line 97: For a small number of tract maps, it may not be a problem.  But I always do `deferLoad=True` to make sure we don't hold too much in memory at once because they all get their own plots.

python/lsst/analysis/tools/tasks/propertyMapTractAnalysis.py
```diff
@@ -0,0 +1,179 @@
...
+            "sum": "Sum",
+        }
+
+        # Making connections for the maps that are configured to run.
+        for propertyName in config.properties:
+            coaddName = config.properties[propertyName].coaddName
+            for operationName in config.properties[propertyName].operations:
+                operationLongName = operationNameLookup[operationName]
+                name = f"{coaddName}Coadd_{propertyName}_map_{operationName}"
+                setattr(
```
erykoff, line 88: Does this actually work for the connections?  Because before one of Jim's refactoring it didn't.  But if it does now, that's great!

erykoff (commented):
A bunch of small comments here, and one big one: how is this tested?

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+                sp = skyproj.GnomonicSkyproj(
+                    ax=ax,
+                    lon_0=tract.ctr_coord.getRa().asDegrees(),
+                    lat_0=tract.ctr_coord.getDec().asDegrees(),
+                    extent=extent,
+                    rcparams=rcparams,
+                )
+                sp.draw_hspmap(mapData, zoom=zoom)
+                sp.set_xlabel("RA (deg)")
+                sp.set_ylabel("Dec (deg)")
```
erykoff, line 212: Oh another thing: `(deg)` is redundant because the tick labels all have the units.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -0,0 +1,443 @@
...
+            values = mapData[mapData.valid_pixels]
+            goodValues = np.isfinite(values)
+            values = values[goodValues]  # As a precaution.
+
+            # Make a concise human-readable label for the plot.
+            plotInfo["coaddName"] = mapName.split("Coadd_")[0]
+            plotInfo["operation"] = self.getLongestSuffixMatch(
+                mapName, ["min", "max", "mean", "weighted_mean", "sum"]
+            )
+            plotInfo["property"] = mapName[
```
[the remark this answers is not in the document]
enourbakhsh: Same as above.

enourbakhsh:
> A bunch of small comments here, and one big one: how is this tested?

I've incorporated a `plotPropertyMapTractExtended` task into `analysis_tools/pipelines/coaddQualityExtended.yaml`. It'll be executed as part of `ci_hsc` via [this line](https://github.com/lsst/drp_pipe/blob/23615f0d050231d0e317c8e0bbb43d3de2505b95/pipelines/HSC/DRP-ci_hsc.yaml#L17). Does that cover it?

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
enourbakhsh: However, I'm uncertain if it might lead to issues in particular scenarios, noticeably altering the aspect ratios of the zoomed-in plots compared to the full-tract plot.

erykoff: I'm pretty sure that _not_ doing this scaling will mess up the aspect ratio.  The full tract is defined in terms of degrees and will have the scaling naturally included.  So that's why we need it here for the zoom as well.

python/lsst/analysis/tools/tasks/propertyMapTractAnalysis.py
erykoff: Well, you were the one who wrote the validation code, so that can be fixed.  But I suppose you would also want to check the type contained in the deferred dataset handle.  This can be done with `.storageClass`.  But at the same time I don't know if using deferred handles is some sort of analysis tools anti-pattern.

erykoff (approved):

python/lsst/analysis/tools/tasks/propertyMapTractAnalysis.py
enourbakhsh: Done!

enourbakhsh:
> > A bunch of small comments here, and one big one: how is this tested?
> 
> I've incorporated a `plotPropertyMapTractExtended` task into `analysis_tools/pipelines/coaddQualityExtended.yaml`. It'll be executed as part of `ci_hsc` via [this line](https://github.com/lsst/drp_pipe/blob/23615f0d050231d0e317c8e0bbb43d3de2505b95/pipelines/HSC/DRP-ci_hsc.yaml#L17). Does that cover it?

Now, all the maps are housed in the coaddQualityCore.yaml pipeline. They are set to undergo testing in ci_hsc via `$DRP_PIPE_DIR/pipelines/_ingredients/HSC/DRP.yaml` found in `$DRP_PIPE_DIR/pipelines/HSC/DRP-ci_hsc.yaml`.

PR #96, opened 2023-05-08
DM-39089: Handle failure modes in Sasquatch Dispatch

natelust:
Handle failure modes in Sasquatch Dispatching where the reference_package is a None value, and when there is no data in a bundle to upload.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
+++ b/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
@@ -354,7 +354,8 @@ class SasquatchDispatcher:
                 except ValueError:
                     # Could not extract package timestamp leaving empty
                     pass
-        meta["reference_package"] = ref_package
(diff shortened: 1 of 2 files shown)


python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
```diff
@@ -354,7 +354,8 @@ def _handleReferencePackage(self, meta: MutableMapping, bundle: MetricMeasuremen
                 except ValueError:
                     # Could not extract package timestamp leaving empty
                     pass
-        meta["reference_package"] = ref_package
+        # explicit handle if None was set in the bundle for the package
+        meta["reference_package"] = ref_package or ""
```
afausti, line 358: Looks good, fields must be present in the payload even if the value is missing. An empty string is a good default value.

afausti (approved):
All changes look good.

PR #103, opened 2023-05-10
DM-39131: Avoid setting infinite limits in scatterPlot

taranu:


python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -173,45 +179,41 @@ class ScatterPlotWithTwoHists(PlotAction):
...
+            if datatype in self.plotTypes:  # type: ignore
+                base.append((f"x{defaults.suffix_xy}", Vector))
+                base.append((f"y{defaults.suffix_xy}", Vector))
+                base.append((f"{datatype}HighSNMask", Vector))
+                base.append((f"{datatype}LowSNMask", Vector))
+                # statistics
+                for name in self._stats:
+                    base.append((f"{{band}}_highSN{defaults.suffix_stat}_{name}", Scalar))
+                    base.append((f"{{band}}_lowSN{defaults.suffix_stat}_{name}", Scalar))
         base.append(("lowSnThreshold", Scalar))
```
sr525, line 217: This is great

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -369,90 +371,38 @@ def _scatterPlot(
...
-                    data["anyLowSNMask"],
-                    "purple",
-                    None,
-                    highStats,
-                    lowStats,
+
+        colormaps_datatype = {
+            "stars": newBlues,
+            "galaxies": newReds,
+        }
```
sr525, line 378: There should be colour maps for the other types, even if they are just set to None for now and fixed later.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -485,14 +435,18 @@ def _scatterPlot(
...
 
-            [xs1, xs25, xs50, xs75, xs95, xs97] = np.nanpercentile(xs, [1, 25, 50, 75, 95, 97])
+            # Chop off 1/3% from only the finite xs values
+            # (there may be +/-np.Inf values)
+            # TODO: This should be configurable
+            # (both the percentiles and just manually specifying limits)
+            xs1, xs97 = np.nanpercentile(xs[np.isfinite(xs)], (1, 97))
             xScale = (xs97 - xs1) / 20.0  # This is ~5% of the data range
 
             # 40 was used as the default number of bins because it looked good
```
sr525, line 445: Is this comment relevant anymore?

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -732,20 +686,22 @@ def _makeTopHistogram(
         if "galaxies" in self.plotTypes:  # type: ignore
             totalX.append(cast(Vector, data["xGalaxies"]))
         if "unknown" in self.plotTypes:  # type: ignore
-            totalX.append(cast(Vector, data["xUknown"]))
+            totalX.append(cast(Vector, data["xUnknown"]))
         if "any" in self.plotTypes:  # type: ignore
             totalX.append(cast(Vector, data["x"]))
 
+        # This filters out nans, but that probably should have happened earlier
```
sr525, line 693: This should have happened earlier, I don't like taking NaNs out of x and not their corresponding y.

tests/data/test_scatterPlot_lines/line_0.txt
```diff
@@ -1,33 +1,38 @@
-2.165581377360398818e+01 2.537420191914861789e+00
```
sr525, line 1: These tests are really annoying but I think I will make a future ticket to sort out testing in atools in general.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
taranu: Isn't it? So much so that I added similar loop roll-ups for the marginal histograms. Also, it now loops over self.plotTypes rather than looping over all possible plot types and checking if they're in self.plotTypes, so you can set the order of plotting however you like.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
taranu: I added the colormap as an attribute but couldn't find colormaps for unknown or all/any. There is a divergent one with three colors in skyPlot.py.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
taranu: It's fine because it's just for histograms. In fact all of the plotting calls will filter out non-finite points, so the `[x for x in totalX if x == x]` lines were unnecessary. The only issues come with trying to set limits without filtering non-finite points (not just nan ones).

I do agree that filtering should happen earlier. Probably most plots should select S/N > 3 or 5, like the scatterplot unit test does.

tests/data/test_scatterPlot_lines/line_0.txt
taranu: Nah, it's great.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -485,15 +427,23 @@ def _scatterPlot(
...
+                # TODO: This should be configurable
+                # but is there a good way to avoid redundant config params
+                # without using slightly annoying subconfigs?
+                xs1, xs97 = np.nanpercentile(xs[np.isfinite(xs)], (1, 97))
+                xScale = (xs97 - xs1) / 20.0  # This is ~5% of the data range
+                xMin, xMax = (xs1 - xScale, xs97 + xScale)
+                print(f"{xLims=}")
+                xLims[0] = min(xLims[0], xMin)
+                xLims[1] = max(xLims[1], xMax)
+                print(f"{xLims=}")
```
sr525, line 444: Do we want these print statements here?

taranu: Whoops... no, definitely not.

PR #106, opened 2023-05-11
DM-37053: Add AMx/AFx/ADx astrometry metrics

cmsaunders:


jeffcarlin:
Looks great, Clare! A few minor comments, but overall the ticket looks good to merge.

See the inline comment re: selecting on mags between 17-21.5 -- seems unnecessary since we're already doing a S/N selection?

I had a slight concern that the default +/-1 arcmin annulus width might be too large. But it's probably fine, and it's configurable so can be changed if desired.

In a test run I did on RC2, some of the plots (histograms) didn't look great. The existence of large outliers made the auto-scaling select a large binsize and too-broad x-axis range. Not specific to this ticket, but something to consider for histograms in analysis_tools generally?

cmsaunders:
Thanks for the review, @jeffcarlin! I tried to stick as close as possible to what is currently done in `faro`, so that is why there are both S/N and magnitude cuts. The annulus width also is the same as in `faro`. We can think about changing them, but for the time being, I want to make this as close as possible to what is done in `faro` for continuity. 

For the histogram appearance, I will add that to [DM-39163](https://jira.lsstcorp.org/browse/DM-39163), because I think it will be easiest to do all the plotting changes together.

PR #107, opened 2023-05-15
DM-36208: Add a README to analysis_tools

leeskelvin:


README.rst
```diff
@@ -1,7 +1,9 @@
-##############
+==============
 analysis_tools
-##############
+==============
 
-``analysis_tools`` is a package in the `LSST Science Pipelines <https://pipelines.lsst.io>`_.
+The Analysis Tools package is designed to assist in the creation of quality assurance (QA) plots and metrics from the outputs of a data reduction pipeline.
```
parejkoj, line 5:
```suggestion
The Analysis Tools package is designed to assist in the creation of quality assurance (QA) plots and metrics from the outputs of a PipelineTask.
```

README.rst
```diff
@@ -1,7 +1,9 @@
...
+The intention is that users have the flexibility to construct complex data analysis tasks from a set of simple building blocks.
+In this sense, a series of consistent, repeatable and high-quality plots and metrics can be generated for any given dataset.
```
parejkoj, line 7: "In this sense"? Maybe something like this:

```suggestion
Analysis Tools allows users to generate consistent, repeatable, and high-quality plots and metrics on the output of a PipelineTask, no matter which dataset it is run on.
```

parejkoj (commented):
A couple of suggestions. Thanks for expanding this.

README.rst
leeskelvin: Barring the pipeline task issue above, this is 👍

README.rst
parejkoj: "data reduction pipeline" is either too specific (not just DRP!) or not specific enough (it has to be one of our LSST Pipelines). Maybe "from the outputs of an LSST Pipeline" and link the latter two to `https://pipelines.lsst.io/modules/lsst.pipe.base/creating-a-pipeline.html`? I don't know if we have a better summary page for what we mean when we say "Pipeline", but it's a more specific thing than what non-LSST people mean.

leeskelvin: DRP != Data Reduction Pipeline.
DRP = Data Release Production. 

Does that change your thinking? I think the phrase "data reduction pipeline" is sufficiently instantly recognisable as to what we mean here - namely, a pipeline that has reduced data. 

Happy to add a URL link.

parejkoj: The link helps. We are specifically talking about our capital-P Pipelines, and that makes it clear.

README.rst
```diff
@@ -1,7 +1,9 @@
-##############
+==============
 analysis_tools
-##############
+==============
 
-``analysis_tools`` is a package in the `LSST Science Pipelines <https://pipelines.lsst.io>`_.
+The Analysis Tools package is designed to assist in the creation of quality assurance (QA) plots and metrics from the outputs of an `LSST data reduction pipeline <https://pipelines.lsst.io/modules/lsst.pipe.base/creating-a-pipeline.html>`_.
+The intention is that users have the flexibility to construct complex data analysis tasks from a set of simple building blocks.
+These analysis tools therefore allow users to generate a series of consistent, repeatable and high-quality plots and metrics, regardless of which dataset they are run on.
```
parejkoj, line 7: I don't think the "therefore" is necessary.

```suggestion
Analysis tools allows users to generate a series of consistent, repeatable and high-quality plots and metrics, regardless of which dataset they are run on.
```

leeskelvin: It helps if you read the fully rendered paragraph, I find: https://github.com/lsst/analysis_tools/tree/tickets/DM-36208.

If sentence 1 in the paragraph begins "The Analysis Tools package..." and then sentence 3 begins "Analysis tools...", it's too much repetition of the phrase in quick succession. This is why I initially preferred "In that sense...", as it's clear were still discussing in the context of sentence 1.

leeskelvin: Use of "therefore" is an attempt to break up that repetition in an attempt to satisfy your comment.

leeskelvin:
PS - as Analysis Tools isn't yet linked in the static docs, I've added an additional sentence linking to the current daily build. I think this will be useful in the coming weeks, as in-kind collaborators begin adding code and wanting to reference what we have right now.

PR #108, opened 2023-05-17
DM-37193: Reorganize vectorActions

taranu:


python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -101,128 +110,63 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
-    def __call__(self, data: KeyedData, **kwargs) -> Vector:
-        vecA = self.actionA(data, **kwargs)  # type: ignore
-        vecB = self.actionB(data, **kwargs)  # type: ignore
-        return vecA / vecB
+        return np.array(sn)
 
 
-class LoadVector(VectorAction):
-    """Load and return a Vector from KeyedData."""
+class ConvertFluxToMags(VectorAction):
```
leeskelvin, line 151: Either this should be named `ConvertFluxToMag` or `ConvertFluxesToMags`. I prefer the singular myself.

python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -101,128 +110,63 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
...
+    """Turn nano janskies into magnitudes."""
```
leeskelvin, line 152: "nanojanskies", one word.

python/lsst/analysis/tools/actions/vector/mathActions.py
```diff
@@ -0,0 +1,132 @@
...
+        yield from self.actionA.getInputSchema()  # type: ignore
+        yield from self.actionB.getInputSchema()  # type: ignore
+
+    def __call__(self, data: KeyedData, **kwargs) -> Vector:
+        vecA = self.actionA(data, **kwargs)  # type: ignore
+        vecB = self.actionB(data, **kwargs)  # type: ignore
+        return vecA / vecB
+
+
+class FractionalDifference(VectorAction):
```
leeskelvin, line 119: Most vector actions in the `mathActions` file end with the suffix `Vector` (all except here and `ConstantValue`). Is it worth considering making that consistent here too?

python/lsst/analysis/tools/actions/vector/mathActions.py
```diff
@@ -0,0 +1,132 @@
...
+import numpy as np
+from lsst.pex.config import Field
+from lsst.pex.config.configurableActions import ConfigurableActionField
+
+from ...interfaces import KeyedData, KeyedDataSchema, Vector, VectorAction
+
+_LOG = logging.getLogger(__name__)
+
+
+class ConstantValue(VectorAction):
```
leeskelvin, line 43: Most vector actions in the `mathActions` file end with the suffix `Vector` (all except here and `FractionalDifference`). Is it worth considering making that consistent here too? E.g., `ConstantVector`? `ConstantValueVector`?

taranu: I was actually considering it renaming it to `Scalar`something rather than `Constant`something, but decided against it.

But to the extent that there is a naming scheme, the `Vector`-suffixed actions all begin with verbs (see also `LoadVector` and `DownselectVector` in `vectorActions.py`, whereas the unsuffixed actions are nouns (`ConstantValue`, `MagDiff` as short for `MagnitudeDifference`, etc). In principle you could verbify the various `Diff(erence)` actions since difference is a verb too, but I don't think the names would be any clearer. For `ConstantValue`, I can't think of a good way to verbify it at all.

leeskelvin: Fair enough. In which case, I suggest leaving this (and the other non-conformant action here) as-is for the time being. Further changes can be punted to a future ticket, if desired.

PR #114, opened 2023-05-24
DM-39162: Return empty focalPlanePlot in case of no data

cmsaunders:


python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -170,8 +176,10 @@ def makePlot(
             focalPlane_x[detectorInd] = fp_x
             focalPlane_y[detectorInd] = fp_y
 
-        binsx = np.linspace(focalPlane_x.min(), focalPlane_x.max(), self.nBins)
-        binsy = np.linspace(focalPlane_y.min(), focalPlane_y.max(), self.nBins)
+        # Add an arbitrary small offset to bins to ensure that the minimum does
+        # not equal the maximum.
```
taranu, line 180: When would you get min==max on either axis, besides having only one point (or all values equal)? Wouldn't the plot be rather unhelpful in that case anyway?

taranu (approved):
Looks fine, but I did want to understand the issue I commented on.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: Yeah...I agree that it's unlikely, but I was looking at what was done in SkyPlot to prevent crashes, and this is in there. Maybe this is just a situation where the plot should fail, because you shouldn't be in a situation with only one point or all x and y values equal. However, if the plot doesn't fail, at least you might be able to see that you have an issue from looking at the plot.

taranu: Fair enough, I suppose we can review what to do with unhelpful plots more generally in the future.

PR #115, opened 2023-05-24
Implement testLoadVector first version

nsevilla:
This is to exercise the workflow we started in the pair coding session at the Commissioning Science Validation Bootcamp.

### DIFF
--- a/tests/test_actions.py
+++ b/tests/test_actions.py
@@ -247,7 +247,11 @@ class TestVectorActions(unittest.TestCase):
 
     # Basic vectorActions
 
-    # def testLoadVector(self): TODO: implement
+    def testLoadVector(self):
+        action = LoadVector(vectorKey="vector")
+        inputData = {"vector": np.arange(5)}
+        result = action(inputData)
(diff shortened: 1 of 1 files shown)


tests/test_actions.py
```diff
@@ -247,7 +247,11 @@ def testFractionalDifference(self):
 
     # Basic vectorActions
 
-    # def testLoadVector(self): TODO: implement
+    def testLoadVector(self):
+        action = LoadVector(vectorKey="vector")
+        inputData = {"vector": np.arange(5)}
```
taranu, line 252:
```suggestion
        inputData = {"vector": [0, 1, 2, 3, 4]}
```
Just to illustrate that you can make (or may receive) suggestions on code reviews. I generally incorporate them into existing commits through an interactive rebase, as [recommended in the dev guide](https://developer.lsst.io/work/flow.html#pushing-code).

taranu (changes requested):
This is an example review. Since this is a user branch with code changes, it should not be merged - the actual changes will go on a future DM ticket.

PR #119, opened 2023-05-25
DM-39345: Prefix CalcBinnedStats metric names by band

taranu:
... `if self.parameterizedBand`, which is always True for the only tool using this action at the moment.

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -51,16 +51,31 @@ class TestDiffMatched(TestCase):
             self.assertGreater(len(tester(data)), 0)
 
     def testMatchedRefCoaddMetric(self):
-        tester = MatchedRefCoaddMetric(unit="")
+        kwargs = {key: "" for key in ("unit", "name_prefix")}
+        for kwarg in kwargs:
+            kwargs_init = {kwarg: ""}
+            tester = MatchedRefCoaddMetric(**kwargs_init)
+            # Validation will fail because configureMetrics hasn't been called
+            with self.assertRaises(ValueError):
+                tester.validate()
+            # Failing to find any of the required keys
+            with self.assertRaises(KeyError):
+                tester({})
+        tester = MatchedRefCoaddMetric(**kwargs)
         with self.assertRaises(ValueError):
-            tester({})
-        tester = MatchedRefCoaddMetric(name_prefix="")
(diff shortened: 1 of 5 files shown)


tests/test_metrics.py
```diff
@@ -51,17 +51,32 @@ def _testMatchedRefCoaddMetricDerived(self, type_metric: type[MatchedRefCoaddMet
...
-        tester = MatchedRefCoaddMetric(name_prefix="")
-        with self.assertRaises(ValueError):
-            tester({})
-        tester = MatchedRefCoaddMetric(unit="", name_prefix="")
+            tester.validate()
+
         tester.finalize()
-        self.assertGreater(len(list(tester.getInputSchema())), 0)
+        # This works, although it leaves none keys in the schema
+        # Should maybe be caught in populatePrepFromProcess
```
cmsaunders, line 70: No change needed, but just a comment that, from talking to Nate, I got the impression that it's inevitable that some things might not get set until `finalize`.

cmsaunders (approved):
This looks fine, though the chain of name formatting is pretty difficult to follow. I do think that we are getting into a situation of more and more arcane names in `analysis_tools`, but that's not something to address on this ticket.

tests/test_metrics.py
taranu: The diffMatched tools aren't using finalize yet, but they will be once DM-36446 is merged and I'll review the tests on that ticket (amongst many other things). In this case I made this comment because it's actually subclasses of this tool that set the relevant configs so it's somewhat inevitable that the base class won't quite work correctly on its own.

taranu:
> This looks fine, though the chain of name formatting is pretty difficult to follow. I do think that we are getting into a situation of more and more arcane names in `analysis_tools`, but that's not something to address on this ticket.

Yeah, I thought about this. The naming scheme is `u_photom_mag_cModelFlux_[all/resolved/unresolved]_[diff/chi]_[median/sigmaMad]_ref_mag17`. Out of all that, the `photom_mag` part could arguably be dropped. Those prefixes add some clarity when the metrics are on their own but aren't entirely necessary when they're part of a metric bundle, except in the case of `y_astrom_y_`, which partly helps to disambiguate between the first y (the band) and the second (the spatial coordinate).

PR #124, opened 2023-06-08
DM-39605: Use butler.dimensions rather than butler.registry.dimensions

timj:


leeskelvin (approved):

leeskelvin:
Looks good to me :+1:

PR #128, opened 2023-06-20
DM-39726: Use warnings instead of numpy.warnings

timj:
The numpy.warnings alias has been removed.

### DIFF
--- a/python/lsst/analysis/tools/actions/vector/vectorActions.py
+++ b/python/lsst/analysis/tools/actions/vector/vectorActions.py
@@ -35,6 +35,7 @@ __all__ = (
 )
 
 import logging
+import warnings
 from typing import Optional, cast
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/actions/vector/vectorActions.py
```diff
@@ -144,9 +145,9 @@ def getInputSchema(self) -> KeyedDataSchema:
         return ((self.vectorKey, Vector),)
 
     def __call__(self, data: KeyedData, **kwargs) -> Vector:
-        with np.warnings.catch_warnings():  # type: ignore
-            np.warnings.filterwarnings("ignore", r"invalid value encountered")  # type: ignore
-            np.warnings.filterwarnings("ignore", r"divide by zero")  # type: ignore
+        with warnings.catch_warnings():
```
timj, line 148: @natelust it looks like the mypy configuration does not warn about irrelevant ignores.

erykoff (approved):

PR #130, opened 2023-06-28
DM-39820: Add a GHA to prevent merge from 'main'

arunkannawadi:
This is an example PR that calls the relevant workflow and show that the action correctly fails when there is a merge from main.


arunkannawadi:
I did another 'goof up' where my local and remote branches diverged. I pull in the changes from remote and pushed again. We should catch this as well.

arunkannawadi:
Closing this without merging since K-T has included this in all repos

PR #131, opened 2023-07-10
DM-32450: Add camera geometry plotting utilities

czwa:


python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -230,3 +232,253 @@ def makePlot(
...
+                # reference the maximum edges of the box in U/V
+                # coordinates.
+                baseU = minU if rotation % 4 in (0, 1) else maxU
+                baseV = maxV if rotation % 4 in (2, 3) else minV
+
+                for amplifier in detector:
+                    ampName = amplifier.getName()
+                    detectorInd = data["detector"] == detectorId
+                    ampInd = data["x"] == ampName
+                    detectorInd &= ampInd
```
cmsaunders, line 395: Personally, I think it would be clearer to reverse this to `ampInd &= detectorInd` so that you are using `ampInd` later.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -230,3 +232,253 @@ def makePlot(
         fig = addPlotInfo(fig, plotInfo)
 
         return fig
+
+
+class FocalPlaneGeometryPlot(FocalPlanePlot):
+    """Plots the focal plane distribution of a parameter in camera
+    geometry units.
```
cmsaunders, line 239: Can you clarify that camera geometry units means detectors or amplifiers?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -230,3 +232,253 @@ def makePlot(
...
+                # negative offsets.  This corresponds to corners that
+                # reference the maximum edges of the box in U/V
+                # coordinates.
+                baseU = minU if rotation % 4 in (0, 1) else maxU
+                baseV = maxV if rotation % 4 in (2, 3) else minV
+
+                for amplifier in detector:
+                    ampName = amplifier.getName()
+                    detectorInd = data["detector"] == detectorId
+                    ampInd = data["x"] == ampName
```
cmsaunders, line 394: If `x` is supposed to be the amp name, can you add that to the docs?

cmsaunders (commented):
This looks good--I mostly just have comments about things that might be able to be made simpler. However, can you put some example plots for the two different levels on the jira ticket?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -230,3 +232,253 @@ def makePlot(
...
+    Given the detector positions in x and y, the focal plane positions
+    are calculated using the camera model. A 2d binned statistic
+    (default is mean) is then calculated and plotted for the parameter
+    z as a function of the camera geometry segment the input points
+    fall upon.
+    """
+
+    xAxisLabel = Field[str](doc="Label to use for the x axis.", optional=False)
+    yAxisLabel = Field[str](doc="Label to use for the y axis.", optional=False)
+    zAxisLabel = Field[str](doc="Label to use for the z axis.", optional=False)
```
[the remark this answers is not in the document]
czwa: I believe so.  I've removed them and confirmed that the plots look identical.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -230,3 +232,253 @@ def makePlot(
...
+            minX, minY = detector.getBBox().getMin()
+            maxX, maxY = detector.getBBox().getMax()
+
+            if self.level.lower() == "detector":
+                detectorInd = data["detector"] == detectorId
+
+                # This does the appropriate statistic for this
+                # detector's data.
+                statistic, _, _ = binned_statistic_dd(
+                    [focalPlane_x[detectorInd], focalPlane_y[detectorInd]],
```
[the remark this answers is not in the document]
czwa: The contents of `focalPlane_xy` are arbitrary, and should be ignored in this block.  I also tried to find another function that understood the same `statistic` argument, and tracked through the source to see that the other `binned_statistic` functions make calls to `binned_statistic_dd`, which does the conversion from the string statistic name to the numeric result itself (here: https://github.com/scipy/scipy/blob/v1.11.1/scipy/stats/_binned_statistic.py#L596).

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
czwa: This comment convinced me to switch from using `x` (which as suggested, isn't obvious) to a new `amplifier` field.  I've updated the docs to reflect this.  This also allows the `x` and `y` vectors to be ignored, allowing a single catalog to be plotted using either `FocalPlanePlot` or `FocalPlaneGeometryPlot` (for `level='detector'`).
I wrote a method to manage some further conversions, but it seemed out of scope for this ticket, so I've saved that as a snippet that might be useful in the future.

cmsaunders (approved):
Very minor comment: you could remove "x" and "y" from the columns that are expected in `data` (lines 290-293).

PR #137, opened 2023-07-25
DM-40180: Remove 'all' histograms from scatter plot with two hists

sr525:


python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -672,17 +672,17 @@ def _makeTopHistogram(
...
         # Top histogram
         topHist = figure.add_subplot(gs[0, :-1], sharex=ax)
+        x_min, x_max = ax.get_xlim()
+        bins = np.linspace(x_min, x_max, 100)
 
-        if "any" in self.plotTypes:
-            x_all = data[f"x{self._datatypes['any'].suffix_xy}"]
-            keys_notall = [x for x in self.plotTypes if x != "any"]
+        if "all" in self.plotTypes:
+            x_all = f"x{self._datatypes['all'].suffix_xy}"
```
laurenam, line 679: Mix of " vs '.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -672,17 +672,17 @@ def _makeTopHistogram(
...
+            keys_notall = [x for x in self.plotTypes if x != "all"]
+            topHist.hist(x_all, bins=bins, color="grey", alpha=0.3, log=True, label=f"All ({len(x_all)})")
```
laurenam, line 681: Why are you still defining an `x_all` below the `else`?  Ditto for `y_all` below.  Actually, you can probably move this setting of `keys_notall` above the `if` and just get rid of the `else`?  Same below, again.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -711,17 +711,17 @@ def _makeSideHistogram(
     ) -> None:
         # Side histogram
         sideHist = figure.add_subplot(gs[1:, -1], sharey=ax)
```
laurenam, line 713: Pre-existing, but there is a rogue `# Side histogram` comment at line 701.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
```diff
@@ -672,17 +672,17 @@ def _makeTopHistogram(
     ) -> None:
         # Top histogram
         topHist = figure.add_subplot(gs[0, :-1], sharex=ax)
+        x_min, x_max = ax.get_xlim()
+        bins = np.linspace(x_min, x_max, 100)
 
-        if "any" in self.plotTypes:
-            x_all = data[f"x{self._datatypes['any'].suffix_xy}"]
-            keys_notall = [x for x in self.plotTypes if x != "any"]
+        if "all" in self.plotTypes:
```
laurenam, line 678: I don’t know what `any` actually means in this context, but now the behaviour is that it considered a `notall` key.  Does this still do the “right thing”?

laurenam (approved):

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
sr525: It needs both to have the string key inside the string.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
sr525: The all should probably be any which brings it in line with the key used in the datatype option. It is for if people don't care if it is a star or a galaxy and just want to plot everything.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
sr525: It should all be any, not sure what I was doing here.

python/lsst/analysis/tools/actions/plot/scatterplotWithTwoHists.py
sr525: This comment is just reminding people that they are in the chunk that defines the side histogram.

PR #138, opened 2023-07-25
DM-39332: Fix the PF1 metric to be threshold relative to mean

jeffcarlin:
This ticket makes the PF1 metric calculation ("stellarPhotRepeatOutlierFraction") use a threshold _relative to the median_ as defined in the SRD, rather than taking a simple threshold in rms. It involved changing the ScalarAction and HistPlot.

### DIFF
--- a/python/lsst/analysis/tools/actions/plot/histPlot.py
+++ b/python/lsst/analysis/tools/actions/plot/histPlot.py
@@ -559,7 +559,7 @@ class HistPlot(PlotAction):
             else:
                 reference_value = self.panels[panel].referenceValue
                 reference_label = "${{\\mu_{{ref}}}}$: {}".format(reference_value)
-        ax2.axvline(reference_value, ls="-", lw=1, c="black", zorder=0, label=reference_label)
+            ax2.axvline(reference_value, ls="-", lw=1, c="black", zorder=0, label=reference_label)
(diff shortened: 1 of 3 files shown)


python/lsst/analysis/tools/actions/plot/histPlot.py
```diff
@@ -536,7 +541,7 @@ def _calcStats(self, data):
         mad = sigmaMad(data)
         return num, med, mad
 
-    def _addReferenceLines(self, ax, panel, panel_range, legend_font_size=7):
+    def _addReferenceLines(self, ax, panel, panel_range, meds, legend_font_size=7):
```
sr525, line 544: Should meds be a kwarg? Looks like it is optional later.

jeffcarlin: I think it makes more sense to keep it a required input. Even though it is only used if the `refRelativeToMedian` flag is set, I'd rather be sure it's there when that flag is set than have an error when the flag is set and `meds` is _not_ passed.

jeffcarlin:
Jenkins has successfully built at https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/97/pipeline, so I'm merging.

PR #144, opened 2023-09-13
DM-40770: Enable config-aware output naming in analysis tools

enourbakhsh:


arunkannawadi (commented):
Is there an action that explicitly uses the `config` passed?

enourbakhsh:
> Is there an action that explicitly uses the `config` passed?

Yes, [here](https://github.com/lsst/analysis_tools/blob/47298a82e91a0a7345b6c3b0058dbc191a4fe06d/python/lsst/analysis/tools/actions/plot/propertyMapPlot.py#L182) is an example.

python/lsst/analysis/tools/interfaces/_analysisTools.py
```diff
@@ -234,7 +238,7 @@ def populatePrepFromProcess(self):
         """
         self.prep.addInputSchema(self.process.getInputSchema())
 
-    def getOutputNames(self) -> Iterable[str]:
+    def getOutputNames(self, **kwargs) -> Iterable[str]:
```
arunkannawadi, line 241: `config: pexConfig.Config | None =None` or something similar would be better.

python/lsst/analysis/tools/interfaces/_analysisTools.py
```diff
@@ -251,7 +255,7 @@ def getOutputNames(self) -> Iterable[str]:
             case JointAction(plot=NoPlot()):
                 return tuple()
             case _HasOutputNames():
-                outNames = tuple(self.produce.getOutputNames())
+                outNames = tuple(self.produce.getOutputNames(**kwargs))
```
arunkannawadi, line 258: config=config

arunkannawadi (approved):
Thanks, missed the example since it was not in the PR.

I'd prefer if we called the argument `config` explicitly instead of burying them under `**kwargs`. Either way, some addition to docstrings are required to explain what fields/attributes `config` should have, regardless of how it is passed.

python/lsst/analysis/tools/interfaces/_actions.py
```diff
@@ -251,14 +259,24 @@ def getInputSchema(self) -> KeyedDataSchema:
...
+            needs to be config-aware.
+
         Returns
         -------
         result : `Iterable` of `str`
             If a `PlotAction` produces more than one plot, this should be the
             keys the action will use in the returned `Mapping`.
         """
-        return self.plot.getOutputNames()
+        # Check if the method accepts a 'config' parameter.
```
arunkannawadi, line 278: No, I don't like calling `inspect` here. It's not good to use this in production, and is what is called as "code smell". The `dynamicOutputNames` should tell you whether it accepts `config` or not already.

timj: The other way is something like:

```python
try:
    names = self.plot.getOutputNames(config=config)
except AttributeError:
    names = self.plot.getOutputNames()
return names
```
`inspect` can be very slow.

enourbakhsh: Thanks for the suggestions! I've ditched the `inspect` approach and made the necessary tweaks using `config` itself. While the `try...except` method works, using the value of `config` seems like a more intuitive way:

``` python
if config is None:
    # `dynamicOutputNames` is set to False. 
    names = self.plot.getOutputNames()
else:
    # `dynamicOutputNames` is set to True.
    names = self.plot.getOutputNames(config=config)
return names
```

arunkannawadi: I like checking if `config` is None as much as the try-except solution that Tim posted.

arunkannawadi (approved):

python/lsst/analysis/tools/interfaces/_actions.py
timj: If you know that a defined `config` means that `config=config` will work then the `if` block is fine with me. Another way to write it is:

```python
kwargs = {} if config is None else {"config": config}
return self.plot.getOutputNames(**kwargs)
```

arunkannawadi: Yup. And just to clarify my earlier comment to Erfan, I'd like the function signature to clearly have `config` as a parameter. Passing it in as `**kwargs` is okay!

timj: It would be simpler if `getOutputNames` always allowed `config=None`.

PR #145, opened 2023-09-18
DM-40768: Stellar photometric repeatability in analysis_tools is not applying defaults correctly

jeffcarlin:
This PR implements a fix to ensure that "PF1" (the fraction of outliers in photometric repeatability) is calculated correctly in analysis_tools. The previous implementation did not actually set the value given in config, so the configured values needed to be moved to a `finalize` method to ensure that they are properly assigned.


bechtol (approved):
Verified that configuration is updated as expected with new finalize function.

PR #148, opened 2023-10-04
DM-39745: Merge matched difference plot/metric classes

taranu:


pipelines/coaddDiffMatchedQualityExtended.yaml
```diff
@@ -1,56 +1,68 @@
...
   diff_matched_analysis:
     class: lsst.analysis.tools.tasks.DiffMatchedAnalysisTask
     config:
       connections.outputName: diff_matched_truth_summary_objectTable_tract
 
       # plots
-      atools.matchedRefCModelMagDiff: MatchedRefCoaddDiffMagPlot
+      atools.matchedRefCModelMagDiff: MatchedRefCoaddDiffMagTool
+      # Doesn't work; interpreted as float
+      # atools.matchedRefCModelMagDiff.produce.plot.xLims: [parameters.mag_x_min, parameters.mag_x_max]
```
leeskelvin, line 23: Does this need to be here for any particular historical reason?

python/lsst/analysis/tools/atools/diffMatched.py
```diff
@@ -216,58 +206,53 @@ def finalize(self):
...
+    )
+    coord_ref = Field[str](
+        doc="The key for reference values of the astrometric variabler",
         optional=False,
     )
 
     def finalize(self):
         # Check if it has already been finalized
         if not hasattr(self.process.buildActions, "diff"):
+            # Set before MagnitudeScatterPlot.finalize to undo PSF default
```
leeskelvin, line 249: Missing period at the end of this comment, unless you intend for these two lines to be read as one sentence (I think not).

pipelines/coaddDiffMatchedQualityExtended.yaml
taranu: Sort of; it's a note that this doesn't work even though it should. In this case I ended up preferring to define `lims_mag_x` in the python block. Anyway, I'm going to address this on a subsequent ticket ([DM-41049](https://jira.lsstcorp.org/browse/DM-41049)).

PR #154, opened 2023-10-11
DM-41118: Add metrics from diffim task metadata

isullivan:


python/lsst/analysis/tools/atools/diffimMetadataMetrics.py
```diff
@@ -0,0 +1,221 @@
...
+    "SciencePsfSizeMetric",
+    "TemplatePsfSizeMetric",
+    "ScienceVarianceScaleMetric",
+    "TemplateVarianceScaleMetric",
+    "TemplateCoverageMetric",
+)
+
+from ..actions.scalar import MedianAction
+from ..interfaces import AnalysisTool
+
```
natelust, line 41: There is a lot of duplication here, I'm going to make some suggestions that make use of Analysis_Tools designed configurability.

Start by deleting all of the metrics below and replace with one tool:

```python
# Pick whatever name you want
class DiffimMetadataMetricTool(AnalysisTool):
    """This tool is designed to calculate the median of data produced from diffim metric data"""
    # This is a convenience field to make it easier to set things from the "top level"
    # you could also just reach right in and set things in config as well and leave
    # this out.
    metricKey = Field[str](doc="The metadata field to process, name will be used as the metric name")
    units = Field[str](doc="The units to associate with the metric")

    def setDefaults(self):
        self.process.calculateActions.diffimMedian = MedianAction(vectorKey="")
        self.produce.metric.units = {"diffimMedian": "ct"}

    def finalize(self):
        # This function is implemented to support the convenience fields defined above
        self.process.calculateActions.diffimMedian.vectorKey = self.metricKey
        self.produce.metric.units["diffimMedian"] = self.units
        self.produce.metric.newNames = {"diffimMedian": self.metricKey}
   
```

Alternatively, we could build a tool that takes a configuration of a dict field for all the metrics to produce with one tool

```python
# Pick whatever name you want
class DiffimMetadataMetricTool(AnalysisTool):
    """This tool is designed to calculate the median of data produced from diffim metric data"""
    metrics = DictField[str, str](doc="The metrics to calculate from the task metadata and their units")

    def finalize(self):
        for name, unit in self.metrics.items():
            setattr(self.process.calculateActions, name, MedianAction(vectorKey=name))
        self.produce.metric.units = dict(self.metrics.items())
```

In this case the yaml would change to look something like:
```yaml
atools.diffimMetadataMetric: DiffimMetaDatametricTool
atools.diffimMetadatametric.metrics:
  nMergedDiaSources: ct
  nGoodPixels: ct
  ...
```

This would result in a single atools bundle uploaded to sasquatch with multiple measurements inside it. Both are perfectly reasonable ways to do things, and the dashboards can handle plotting from each the same way. The big difference is that if you are getting these back from the butler, the former will result in n dataset types and n `butler.get` calls. The later will be a single dataset type and a single `butler.get` call with multiple values in it.

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
       atools.simpleSky: SimpleDiaPlot
       python: |
         from lsst.analysis.tools.atools import *
+  diffimTaskCore:
+    class: lsst.analysis.tools.tasks.DiffimDetectorVisitAnalysisTask
+    config:
+      connections.outputName: diffimMetadata
+      atools.nUnmergedDiaSources: NUnmergedDiaSourcesMetric
```
natelust, line 19:
```
atools.nUnmergedDiaSources: DiffimMetadataMetricTool
atools.nUnmergedDiaSources.metricKey: nUnmergedDiaSources
atools.nUnmergedDiaSources.units: ct
```

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
...
+      atools.nMergedDiaSources: NMergedDiaSourcesMetric
```
natelust, line 20: atools.nUnmergedDiaSources: DiffimMetadataMetricTool
atools.nUnmergedDiaSources.metricKey: nMergedDiaSources
atools.nUnmergedDiaSources.units: ct

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
...
+      atools.nGoodPixelsMetric: NGoodPixelsMetric
```
natelust, line 21: atools.nUnmergedDiaSources: DiffimMetadataMetricTool
atools.nUnmergedDiaSources.metricKey: nGoodPixel
atools.nUnmergedDiaSources.units: ct

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
...
+      atools.nBadPixelsMetric: NBadPixelsMetric
```
natelust, line 22: similar to above for each of the tools below

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
...
+      atools.nDetectedNegativeAllPixelsMetric: NPixelsDetectedNegativeMetric
+      atools.nDetectedPositiveGoodPixelsMetric: NBadPixelsDetectedPositiveMetric
+      atools.nDetectedNegativeGoodPixelsMetric: NBadPixelsDetectedNegativeMetric
+      atools.sciencePsfSize: SciencePsfSizeMetric
+      atools.templatePsfSize: TemplatePsfSizeMetric
+      atools.scaleScienceVariance: ScienceVarianceScaleMetric
+      atools.scaleTemplateVariance: TemplateVarianceScaleMetric
+      atools.templateCoverage: TemplateCoverageMetric
+      python: |
+        from lsst.analysis.tools.atools import *
```
natelust, line 33: Probably best practice to import your module here instead of star

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -12,3 +12,22 @@ tasks:
       atools.simpleSky: SimpleDiaPlot
       python: |
         from lsst.analysis.tools.atools import *
```
natelust, line 14: The changes below will make sense after reading the metric definition comments.

python/lsst/analysis/tools/tasks/diffimTaskDetectorVisitAnalysis.py
```diff
@@ -0,0 +1,70 @@
...
+import pandas as pd
+from lsst.pipe.base import connectionTypes as ct
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class DiffimDetectorVisitAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions=("visit", "band", "detector"),
+    defaultTemplates={"coaddName": "goodSeeing", "fakesType": "fakes_"},
```
natelust, line 34: I'm confused on why these templates are here?

natelust (commented):
There are a few suggestions to change things up, what path you choose is up to you.

PR #155, opened 2023-10-12
DM-40633: Ignore many numpy warnings

taranu:
These warnings spam the logs with mostly useless messages. You can tell which action they occur in and it's usually obvious why - divide by zero, nan, etc - but these are typically expected and not necessarily a problem for the task.


python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -276,38 +277,54 @@ class AnalysisPipelineTask(PipelineTask):
...
+                # only add the metrics if there are some
+                if metricAccumulate:
+                    results.metrics[name] = metricAccumulate
+            # Wrap the return struct in a finalizer so that when results is
+            # garbage collected the plots will be closed.
+            # TODO: This finalize step closes all open plots at the conclusion
+            # of a task. When DM-39114 is implemented, this step should not
+            # be required and may be removed.
+            weakref.finalize(results, _plotCloser, *weakrefArgs)
+            return results
```
erykoff, line 327: Return should be out of the with context.

python/lsst/analysis/tools/actions/plot/colorColorFitPlot.py
```diff
@@ -155,11 +155,11 @@ def makePlot(
           * Statistics that are shown on the plot or used by the plotting code:
              * ``approxMagDepth``
                    The approximate magnitude corresponding to the SN cut used.
-             * ``f"{self.plotName}_sigmaMAD"``
+             * ``f"{self.plotName}_nanmedian"``
                    The sigma mad of the distances to the line fit.
```
sr525, line 159: Why has this changed from sigmaMad to median?

python/lsst/analysis/tools/actions/plot/colorColorFitPlot.py
```diff
@@ -155,11 +155,11 @@ def makePlot(
...
              * ``f"{self.identity or ''}_median"``
                    The median of the distances to the line fit.
-             * ``f"{self.identity or ''}_hardwired_sigmaMAD"``
+             * ``f"{self.identity or ''}_hardwired_nanmedian"``
```
sr525, line 162: Same comment here

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -36,7 +36,7 @@
 from scipy.stats import binned_statistic_2d, binned_statistic_dd
 
 from ...interfaces import KeyedData, KeyedDataSchema, PlotAction, Scalar, Vector
-from ...statistics import nansigmaMad
+from ...statistics import nanmedian, nansigmaMad
```
sr525, line 39: I think nanMedian and nanSigmaMad would be easier to read

python/lsst/analysis/tools/actions/plot/colorColorFitPlot.py
taranu: Whoops, that was a find/replace fail I missed correcting.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
taranu: I agree, but I was following the numpy naming convention. Having said that, it might make more sense to not follow numpy so as to distinguish them instead?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -22,100 +43,62 @@
...
-
-    Parameters
-    ----------
-    data : `KeyedData`
-
-    Returns
-    -------
-    count : `Scalar`
-        The number of unique rows in a given column.
-    """
```
sr525, line 113: Why did we get rid of the more extensive docs?

python/lsst/analysis/tools/statistics.py
```diff
@@ -19,14 +19,152 @@
...
+    numpy_invalid_value_log,
+    numpy_invalid_value_log10,
+    numpy_invalid_value_sqrt,
+    numpy_mean_empty,
+)
+
+
+def divide(dividend: Scalar | Vector, divisor: Scalar | Vector) -> Scalar | Vector:
+    """Return dividend/divisor."""
+    with warnings.catch_warnings():
```
sr525, line 60: Can we not catch the issue rather than the warnings? 

if divisor == 0:
    return NaN

Or something? I guess it does the same thing but feels more honest. Also doesn't solve my problem that what I really want is it to also tell me why it has failed. Maybe we just need a metric for each whatever that is number of 0, number of NaN etc.

python/lsst/analysis/tools/statistics.py
```diff
@@ -19,14 +19,152 @@
...
+        warnings.filterwarnings(filterwarnings_action, numpy_divide_zero_log10)
+        warnings.filterwarnings(filterwarnings_action, numpy_invalid_value_log10)
+        result = np.log10(values)
+    return result
+
+
+def nansigmaMad(vector: Vector) -> Scalar:
+    """Return the sigma_MAD of a vector."""
+    return cast(Scalar, sps.median_abs_deviation(vector, scale="normal", nan_policy="omit"))
```
sr525, line 119: Is that the sigma MAD or just the MAD?

python/lsst/analysis/tools/math.py
```diff
@@ -0,0 +1,170 @@
...
+    """Return the natural logarithm of values."""
+    with warnings.catch_warnings():
+        warnings.filterwarnings(filterwarnings_action, numpy_divide_zero_log10)
+        warnings.filterwarnings(filterwarnings_action, numpy_invalid_value_log10)
+        result = np.log10(values)
+    return result
+
+
+def nanSigmaMad(vector: Vector) -> Scalar:
+    """Return the sigma_MAD of a vector."""
```
sr525, line 118: This definitely returns the sigma MAD not just the MAD?

PR #157, opened 2023-10-17
DM-41182: Add basic functionality test for PropertyMapTractAnalysisTask

enourbakhsh:
While I was at it, I fixed typos, tweaked the plot action to work with the test, and made other necessary adjustments throughout.

### DIFF
--- a/tests/test_propertyMapPlot.py
+++ b/tests/test_propertyMapPlot.py
@@ -1,0 +1,356 @@
+# This file is part of analysis_tools.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import os
+import unittest
+
+import healsparse as hsp
+import lsst.utils.tests
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+import skyproj
+from lsst.analysis.tools.atools.propertyMap import PropertyMapTool
+from lsst.analysis.tools.tasks.propertyMapTractAnalysis import (
+    PropertyMapConfig,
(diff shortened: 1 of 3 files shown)


tests/test_propertyMapPlot.py
```diff
@@ -0,0 +1,248 @@
...
+            plotInfo=self.plotInfo,
+        )
+
+        # Previously computed reference hash values for the plots.
+        expectedHashList = [
+            "87235818c562587a6f6b17a233df76a1",
+            "90c68d7c6a70536485cea093b9401038",
+            "04385cbc6ead899d210fcaa628f88e45",
+            "b66ca03562bfcfcd758461c5ccb73837",
+            "98f145f450001642b2f1538c629f683f",
```
erykoff, line 154: I'd be very shocked if this works; and if it works today it certainly won't tomorrow.  Matplotlib images differ subtly from mac to linux and different os and matplotlib versions.  You can either use something like `test_scatterPlot.py` or check in sample images to the tests and use the matplotlib image comparison testing routines which allow some amount of variance from plotted image to reference image.

tests/test_propertyMapPlot.py
```diff
@@ -0,0 +1,248 @@
...
+        task = PropertyMapTractAnalysisTask()
+        self.plotConfig = config
+        self.plotInfo = task.parsePlotInfo(mapsDict, dataId, list(mapsDict.keys()))
+        self.data = {"maps": mapsDict}
+        self.atool = PropertyMapTool()
+        self.atool.produce.plot.plotName = "test"
+        self.atool.finalize()
+
+    def tearDown(self):
+        removeTestTempDir(self.testDir)
```
erykoff, line 131: I think this needs to go at the end of the teardown.

erykoff (changes requested):
Image comparison is tricky.  But you can't use a hash and expect it'll match with different os's and versions of everything.  See my comment below.

tests/test_propertyMapPlot.py
```diff
@@ -0,0 +1,356 @@
...
+
+        # Calculate fractions.
+        rActualFraction = np.sum(data[:, :, 0]) / data.size
+        gActualFraction = np.sum(data[:, :, 1]) / data.size
+        bActualFraction = np.sum(data[:, :, 2]) / data.size
+
+        # Check if the actual fractions meet the expected fractions within the
+        # given tolerance.
+        errors = []
+        if not np.abs(rActualFraction - rFraction) <= rtol:
```
erykoff, line 328: This is fine, but we usually use `self.assertFloatsAlmostEqual` which allows an rtol and/or an atol.  I suppose the advantage here is that you get more information on all the channels rather than the first failure, but I wanted to point out the standard tools.

erykoff (approved):
This looks fine, though a bit non-standard in the tests.

tests/test_propertyMapPlot.py
enourbakhsh: Thanks for the heads-up. Correct, I went this route because seeing info on all errors is quite handy for quick debugging, especially when multiple criteria are being checked in one go.

PR #159, opened 2023-10-22
DM-41027: Fix potential infinite loop in analysis_drp colorColorFitPlot

laurenam:


python/lsst/analysis/tools/actions/plot/colorColorFitPlot.py
```diff
@@ -58,6 +58,10 @@ class ColorColorFitPlot(PlotAction):
     )
 
     plotName = Field[str](doc="The name for the plot.", optional=False)
+    minPointsForFit = Field[int](
+        doc="Minimum number of valid objects to bother attempting a fit.",
+        default=5,
```
taranu, line 63: As in `analysis_drp`, should the value be >0?

laurenam: Good idea. I made it a `RangeField` and gave it `min=1`.

PR #160, opened 2023-10-23
DM-41294: DM-41182 breaks nightly build with matplotlib 3.7.2

enourbakhsh:


tests/test_propertyMapPlot.py
```diff
@@ -180,7 +180,7 @@ def test_PropertyMapTractAnalysisTask(self):
             self._validateFigureStructure(fig, binsCount, xlabel, zoomFactors)
 
             # Validate the RGB fractions of the figure.
-            self._validateRGBFractions(fig, expectedRGBFraction, rtol=1e-7)
+            self._validateRGBFractions(fig, expectedRGBFraction, rtol=1e-3)
```
ktlim, line 183: Is there a principled reason for picking this value, or is it just empirical?  A comment with the answer might be good.

ktlim (approved):

tests/test_propertyMapPlot.py
enourbakhsh: It is empirical. Added in the comment.

PR #161, opened 2023-11-03
DM-40519: Add photometry sky plot to coaddQualityCore

ctslater:


python/lsst/analysis/tools/atools/photometry.py
```diff
@@ -89,6 +89,57 @@ def setDefaults(self):
         }
 
 
+class PsfCModelSkyPlot(AnalysisTool):
```
ctslater, line 92: Would be nice to add a one-line docstring; one can pretty much figure it out from the class name but it's nice to have it in text.

ctslater (commented):
All looks good! Two minor comments. Github won't let me approve since I opened the PR...

PR #162, opened 2023-11-06
DM-41598: Add missing abstract methods to SasquatchDatastore

andy-slac:
Also adds a trivial unit test for SasquatchDatastore to verify that it can be instantiated and simple `Butler.put` works on it.


timj (approved):
Looks great. Thanks for adding test.

PR #165, opened 2023-11-21
DM-41623: Enabled the plotting of patch boundaries in skyPlot.py

jrmullaney:


python/lsst/analysis/tools/actions/plot/skyPlot.py
```diff
@@ -211,8 +214,11 @@ def makePlot(
         fig = plt.figure(dpi=300)
         ax = fig.add_subplot(111)
 
-        if sumStats is None:
-            sumStats = {}
+        if self.plotOutlines:
+            if sumStats is None and "patch" in data.keys():
+                sumStats = generateSummaryStats(data, kwargs["skymap"], plotInfo)
+            else:
+                sumStats = {}
```
mfisherlevine, line 221: This doesn't maintain the previous behaviour: if `self.plotOutlines` is `False` then the `None` doesn't get converted to an empty dict like it used to. I think you need to do something more like this:

```
if sumStats is None:
    if self.plotOutlines and "patch" in data.keys():
        sumStats = generateSummaryStats(data, kwargs["skymap"], plotInfo)
    else:
        sumStats = {}
```

mfisherlevine: (untested suggestion, but hopefully that makes sense and will more or less work)

jrmullaney: Agreed. I've changed it to what you suggested, and tested it by running coaddQualityExtended.yaml both with plotOutlines=False and plotOutlines=True. Works as expected under both circumstances. Just about to push new version...

PR #166, opened 2023-11-21
DM-41624: Allow extreme scatter to be optional in skyPlot.py

jrmullaney:


python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -602,6 +602,7 @@ def plotProjectionWithBinning(
     isSorted=False,
     vmin=None,
     vmax=None,
+    addExtremeScatter=True,
```
ctslater, line 605: `addExtremeScatter` is not my favorite name for this, I think of it as something like "show outlier points" or "display large outliers" or some other combination with a more display-oriented verb.  ("add" sounds like it's changing the values)

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -634,6 +635,9 @@ def plotProjectionWithBinning(
         extrema of the data).
     vmin, vmax : `float`, optional
         The min and max limits for the colorbar.
+    addExtremeScatter: `bool`, optional
+        Whether to include scatter points showing the positions of the
+        most extreme 15% head and tail of points?
```
ctslater, line 640: I'd drop the question mark and "Whether to". Points-> values? Also, subtly, it's the 15% most extreme, but one could read this as saying it's the head 15% _and_ the tail 15%.

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -634,6 +635,9 @@ def plotProjectionWithBinning(
         extrema of the data).
     vmin, vmax : `float`, optional
         The min and max limits for the colorbar.
+    addExtremeScatter: `bool`, optional
```
ctslater, line 638: `default True` instead of `optional` is a little more useful.

python/lsst/analysis/tools/actions/plot/skyPlot.py
```diff
@@ -79,6 +79,11 @@ class SkyPlot(PlotAction):
         default=Med2Mad,
     )
 
+    addExtremeScatter = Field[bool](
+        doc="Add extreme scatter points?",
```
ctslater, line 83: I'd use a display-oriented verb here too instead of "add".

ctslater (commented):
Structure-wise everything is good, but I'd see if there's a clearer name that could be used for this parameter. Naming things is the hardest problem!

python/lsst/analysis/tools/actions/plot/plotUtils.py
jrmullaney: ok - proposed name change to "showExtremeOutliers"

ctslater (approved):
Thanks for the changes, looks good!

PR #168, opened 2023-12-01
DM-37727-fix: Make sure that the matching works for imsim as well as HSC

sr525:


yalsayyad:
This commit is terrible, but file a follow ticket to clean up the aftermath, and we can deal with it next week.

PR #169, opened 2023-12-05
DM-41934 Add the ability to explicitly set Sasquatch time

natelust:
Add the ability to explicitly set the time Sasquatch will use from the pipeline task or pipeline interfaces. This is done by setting the task config variable timestamp_version, or the pipeline parameter sasquatch_timestamp_version to a value of
explicit_timestamp:%Y%m%dT%H%M%S%z


python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -212,7 +213,8 @@ class AnalysisBaseConfig(PipelineTaskConfig, pipelineConnections=AnalysisBaseCon
         doc="Which time stamp should be used as the reference timestamp for a "
         "metric in a time series database, valid values are; "
         "reference_package_timestamp, run_timestamp, current_timestamp, "
-        "and dataset_timestamp",
+        "dataset_timestamp and explicit_timestamp:datetime where datetime is"
```
leeskelvin, line 216: Repeated 'is' on the line below. Make sure not to forget the trailing whitespace on correction.

python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -232,6 +234,18 @@ def applyConfigOverrides(
         if (value := parameters.mapping.get("sasquatch_reference_package", None)) is not None:
             extraConfig["reference_package"] = value
         if (value := parameters.mapping.get("sasquatch_timestamp_version", None)) is not None:
+            if "explicit_timestamp" in value:
+                try:
+                    _, splitTime = value.split(":")
+                except ValueError as excpt:
+                    raise ValueError("Could not find time specified with explicit_timestamp") from excpt
```
leeskelvin, line 241: I think this `raise` would be more useful if it read something like:
```
Explicit timestamp must be given in the format 'explicit_timestamp:datetime', where datetime is given in the form '%Y%m%dT%H%M%S%z'
```

python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -212,7 +213,8 @@ class AnalysisBaseConfig(PipelineTaskConfig, pipelineConnections=AnalysisBaseCon
         doc="Which time stamp should be used as the reference timestamp for a "
         "metric in a time series database, valid values are; "
         "reference_package_timestamp, run_timestamp, current_timestamp, "
-        "and dataset_timestamp",
+        "dataset_timestamp and explicit_timestamp:datetime where datetime is"
+        "is given in the form %Y%m%dT%H%M%S%z",
```
leeskelvin, line 217: So, %z is the time zone in format "±HHMM". AFAICT, times uploaded to the Chronograf are in Zulu time, i.e., UTC+0, i.e., GMT. Would that equate to +0000 in this format?

python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
```diff
@@ -408,6 +408,12 @@ def _handleTimes(self, meta: MutableMapping, bundle: MetricMeasurementBundle, ru
                 case "dataset_timestamp":
                     log.error("dataset timestamps are not yet supported, run_timestamp will be used")
                     meta["timestamp"] = meta["run_timestamp"]
+                case str(value) if "explicit_timestamp" in value:
+                    try:
+                        _, splitTime = value.split(":")
+                    except ValueError as excpt:
+                        raise ValueError("Could not find time specified with explicit_timestamp") from excpt
```
leeskelvin, line 415: A similar comment here about the `raise`.

tests/test_sasquatchDatastore.py
```diff
@@ -71,6 +72,24 @@ def test_put(self):
...
+        bundle.timestamp_version = "explicit_timestamp"
+        # verify this raises with no specified time
+        with self.assertRaises(ValueError):
+            dispatcher._handleTimes({}, bundle, "localRun")
+        # verify this raise with a date that can't be parsed
+        bundle.timestamp_version = "explicit_timestamp:123233"
+        with self.assertRaises(ValueError):
+            dispatcher._handleTimes({}, bundle, "localRun")
+        # verify that a correct time gets parsed
+        bundle.timestamp_version = "explicit_timestamp:20230728T165102Z"
```
leeskelvin, line 87: I was confused as to why this works, as the [docs](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes) seem to suggest that this should be `+0000`. However, I've tested it, and `strptime` in a notebook _does_ parse `Z` to `tzinfo=datetime.timezone.utc`. It doesn't parse any other time zone code (I.e., any other letter of the alphabet). Is `Z` just a special case?

python/lsst/analysis/tools/interfaces/_task.py
natelust: You can do +0000 or z which is an explicit carve out, I think because it is such a commonly output format

tests/test_sasquatchDatastore.py
natelust: It is an explicit carve out: https://docs.python.org/3.11/library/datetime.html#technical-detail . This has been true formally since python 3.7 and reading various prs and such I think it implicitly worked prior to that and they just formalized it. Because it is the formal regex of the python datetime library, it will not change, as that would break so much backwards compatibility.

PR #172, opened 2023-12-08
DM-41953: Implement an adaptive binning option for projection plots in analysis_tools

jrmullaney:


sr525:
I think it would be useful to have something that defaults to points rather than binning in the case of low density. I wonder if it is possible to do both in an efficent manner on the same plot, probably future work.

jrmullaney:
> I think it would be useful to have something that defaults to points rather than binning in the case of low density. I wonder if it is possible to do both in an efficent manner on the same plot, probably future work.

Thanks @sr525 for taking a look at this. To clarify - are you ok for me to merge this, then create a new ticket to implement your suggestion for plotting points for low densities? Is this what you meant by "future work"?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -54,7 +54,10 @@ class FocalPlanePlot(PlotAction):
     zAxisLabel = Field[str](doc="Label to use for the z axis.", optional=False)
 
     nBins = Field[int](
-        doc="Number of bins to use within the effective plot ranges along the spatial directions.",
+        doc="Number of bins to use within the effective plot ranges along"
+        " the spatial directions. If nBins is negative the number of"
+        " bins is adapted to the source density, with lower densities"
+        " using fewer bins, but with a minimum of abs(nBins).",
```
sr525, line 60: I think that instead of using negative it would be better to have a config option, something like doUseAdaptiveBinning, that is more explicit and easier to understand at a glance. Then the normal bins config can stay the same and be used for both.

PR #173, opened 2023-12-08
DM-41938: Generalise matching tasks to be more useful outside of a reference catalog context

sr525:


python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -125,10 +125,38 @@ class CatalogMatchConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Catalo
...
+
+    refDecColumn = pexConfig.Field[str](
+        doc="Dec column name for the reference (being matched to) catalog.",
+        default="dec",
+    )
+
     patchColumn = pexConfig.Field[str](doc="Patch column.", default="patch")
 
+    refCat = pexConfig.Field[bool](
+        doc="Is the catalog being matched to a reference catalog?",
```
taranu, line 151: I'm confused as to what this actually means. The flag only seems to rename reference magnitude columns if enabled. So should it be disabled when the refCatalog (the one being matched to) doesn't have magnitudes at all (or only fluxes? or neither?), or the naming scheme doesn't match expectations? When would any of those conditions be the case?

Separately, when you say "reference catalog" do you specifically mean a reference catalog used for calibration distinctly from the `refCatalog` being matched?

taranu (commented):
The code seems fine. I have naming suggestions and questions for clarification.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -167,6 +168,17 @@ def setDefaults(self):
...
 
+class MatchingFlagSelector(CoaddPlotFlagSelector):
+    """The default flag selector to apply pre matching.
+    The sources are cut down to remove duplicates but
+    not on quality.
+    """
+
+    def setDefaults(self):
+        self.selectWhenFalse = []
+        self.selectWhenTrue = ["detect_isPrimary"]
```
taranu, line 179: Shouldn't this have `detect_isTractInner` for coadds?

python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -125,10 +125,38 @@ class CatalogMatchConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Catalo
...
+    )
+
+    refDecColumn = pexConfig.Field[str](
+        doc="Dec column name for the reference (being matched to) catalog.",
+        default="dec",
+    )
+
     patchColumn = pexConfig.Field[str](doc="Patch column.", default="patch")
 
+    refCat = pexConfig.Field[bool](
```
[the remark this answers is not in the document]
sr525: I am going to go with matchesRefCat

python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -203,23 +231,43 @@ def run(self, *, catalog, loadedRefCat, bands):
...
         for col in refCols:
-            loadedRefCatMatched.rename_column(col, col + "_ref")
-
-        for i, band in enumerate(bands):
-            loadedRefCatMatched[band + "_mag_ref"] = loadedRefCatMatched["refMag_ref"][:, i]
-            loadedRefCatMatched[band + "_magErr_ref"] = loadedRefCatMatched["refMagErr_ref"][:, i]
-        loadedRefCatMatched.remove_column("refMag_ref")
-        loadedRefCatMatched.remove_column("refMagErr_ref")
-        tMatched = hstack([targetCatalogMatched, loadedRefCatMatched], join_type="exact")
+            refCatalogMatched.rename_column(col, col + "_ref")
```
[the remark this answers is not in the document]
sr525: Noted but this syntax was inherited from previous work and is in use elsewhere.

python/lsst/analysis/tools/tasks/catalogMatch.py
sr525: It wants to know if the catalog being matched to is stored as a reference catalog because then the fluxes need to be flattened out.

python/lsst/analysis/tools/actions/vector/selectors.py
laurenam: Not when working per-tract as there is no chance of duplication if there's no tract overlap.  Setting it in a per-tract context will unnecessarily leave an empty boundary around the tract (and will almost certainly puzzle people as to why!)  If plotting data with overlapping tracts, then it is best just to use `isPrimary`.

sr525: I thought isPrimary included isTractInner?

laurenam: I rather meant to not change the selection in this ticket (except for excluding sky objects, perhaps) as it will be difficult to identify the change which is unrelated to the ticket when we are trying to figure out why the metrics changed…

PR #178, opened 2023-12-19
DM-40136: Add AA1 metric

cmsaunders:


python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -153,6 +153,7 @@ class FracThreshold(ScalarFromVectorAction):
     threshold = Field[float](doc="Threshold to apply.")
     percent = Field[bool](doc="Express result as percentage", default=False)
     relative_to_median = Field[bool](doc="Calculate threshold relative to " "the median?", default=False)
+    absolute_value = Field[bool](doc="Take absolute value of distribution")
```
jeffcarlin, line 156: Should this have a default? (i.e., `default=False`?)

python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -529,3 +533,73 @@ def setDefaults(self):
...
+        self.process.buildActions.mags = ConvertFluxToMag()
+        self.process.buildActions.mags.vectorKey = "psfFlux_target"
+
+        self.process.filterActions.brightStars = DownselectVector(vectorKey="astromDiff")
+        self.process.filterActions.brightStars.selector = RangeSelector(
+            vectorKey="mags", minimum=17, maximum=21.5
+        )
+
+        self.process.calculateActions.AA1 = MedianAction(vectorKey="brightStars")
+        self.process.calculateActions.AB1 = SigmaMadAction(vectorKey="brightStars")
```
jeffcarlin, line 571: I don't think this is what AB1 is meant to be. It should be the difference between separations measured in the r-band and those measured in any other filter. This seems to be calculating only the RMS of the separations in the input visit's filter?

python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -529,3 +533,73 @@ def setDefaults(self):
...
+            self.process.buildActions.astromDiff.buildAction.actionB = LoadVector(vectorKey="dec_ref")
+
+        self.process.calculateActions.ABF1 = FracThreshold(
+            vectorKey="brightStars",
+            threshold=self.AB2,
+            percent=True,
+            op="gt",
+            relative_to_median=True,
+            absolute_value=True,
+        )
```
jeffcarlin, line 599: I also think this should be calculated using separations in the visit's band relative to separations in the r-band. (i.e., it requires a pair of visits -- one in a non-r band, and one in r-band)

jeffcarlin (commented):
Would it make sense to also calculate a version of AA1 that is the total positional offset? That could be useful to have sometimes.

I place a few comments here, but I think we should discuss the implementation of AB1 and ABF1 on the Jira ticket. I'll post questions there.

python/lsst/analysis/tools/atools/refCatMatchPlots.py
cmsaunders: Renaming AB1 to AA1_sigmaMad.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
cmsaunders: Yes, but I'm moving it to DM-39256 anyway.

python/lsst/analysis/tools/atools/refCatMatchPlots.py
cmsaunders: Moved to DM-39256.

python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -529,3 +534,111 @@ def setDefaults(self):
...
+            "AA1_RA": "{band}_AA1_RA_coadd",
+            "AA1_sigmaMad_RA": "{band}_AA1_sigmaMad_RA_coadd",
+            "AA1_Dec": "{band}_AA1_Dec_coadd",
+            "AA1_sigmaMad_Dec": "{band}_AA1_sigmaMad_Dec_coadd",
+            "AA1_tot": "{band}_AA1_tot_coadd",
+            "AA1_sigmaMad_tot": "{band}_AA1_sigmaMad_tot_coadd",
+        }
+
+    def visitContext(self) -> None:
+        """Apply visit options for the metrics. Applies the coadd plot flag
```
jeffcarlin, line 569: "coadd plot flag" --> "visit plot flag"?

python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -529,3 +534,111 @@ def setDefaults(self):
 
         self.produce.plotName = "astromDiffSky_Dec"
         self.produce.zAxisLabel = "Dec$_{{target}}$ - Dec$_{{ref}}$ (marcsec)"
+
+
+class TargetRefCatDeltaMetrics(AnalysisTool):
+    """Calculate the AA1, AB1, and ABF1 metrics from the difference between the
```
jeffcarlin, line 540: I think it's only calculating AA1 now?

jeffcarlin (approved):
Looks good -- approved!

PR #181, opened 2024-01-10
DM-42427: CountAction result should be an int not a float

isullivan:


erykoff (approved):
Should the other code paths explicitly cast to int just in case?

PR #185, opened 2024-01-17
DM-42535: Add the old config names back in with a deprecation warning.

sr525:


python/lsst/analysis/tools/tasks/catalogMatch.py
```diff
@@ -156,6 +156,18 @@ class CatalogMatchConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Catalo
         default="dec",
     )
 
+    raColumn = pexConfig.Field[str](
+        doc="RA column.",
+        default="coord_ra",
+        deprecated="This field was replaced with targetRaColumn. Will be removed after v27.",
```
erykoff, line 162: I would add "with targetRaColumn and is unused. Will..."

erykoff (approved):
Minor wording comment (applies to both).

PR #190, opened 2024-01-24
DM-42642: Introduce MatrixPlot plot type to analysis_tools

enourbakhsh:


python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,348 @@
...
+                # Set positions for minor ticks at the midpoints of the major
+                # ticks.
+                axis.set_ticks((major_tick_positions[:-1] + major_tick_positions[1:]) / 2, minor=True)
+                # Set minor tick labels.
+                axis.set_ticklabels(major_tick_labels, minor=True)
+                # Remove minor tick marks for asthetic reasons.
+                axis.set_tick_params(which="minor", length=0)
+        else:
+            # Set the desired tick values and labels if provided.
+            if self.xAxisTickValues is not None:
```
taranu, line 310: Just noting that If it's left like this instead of `if self.xAxisTickValues`, then setting `self.xAxisTickValues=[]` (instead of None) will remove all ticks. I think it's ok to support remove axis ticks and labels but maybe there should be a note about this in the relevant config field docs.

taranu (commented):
Looks mostly fine; see specific comments.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,384 @@
...
+from ...interfaces import PlotAction, Vector
+from .plotUtils import addPlotInfo
+
+if TYPE_CHECKING:
+    from matplotlib.figure import Figure
+
+    from ...interfaces import KeyedData, KeyedDataSchema
+
+
+class guideLinesConfig(Config):
```
taranu, line 44: Needs to be `GuideLinesConfig`.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,384 @@
...
+        default=12,
+    )
+
+    xAxisTickValues = ListField[float](
+        doc="List of x-axis tick values. If not set, the ticks will be set automatically by matplotlib.",
+        default=None,
+    )
+
+    xAxisTickLabels = DictField[float, str](
+        doc="Dictionary of x-axis tick positions and labels. These can work independent of "
```
taranu, line 121: These are applied as minor ticks, right? Maybe add a longer note about that in the class docstring, especially if it's possible to misconfigure them such that the major and minor tick labels overlap.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,384 @@
...
+        "`xAxisTickValues`.",
```
taranu, line 122: I'm not sure if this reference is going to be parsed in the docs.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        matrix = data[self.matrixKey]
+        # Calculate default limits only if needed.
+        if self.vmin is None or self.vmax is None:
+            default_limits = apViz.PercentileInterval(98.0).get_limits(np.abs(matrix.flatten()))
+        else:
+            default_limits = (None, None)
+
+        # Set the value range using overrides or defaults.
+        vrange = (self.vmin or default_limits[0], self.vmax or default_limits[1])
```
taranu, line 222: I think you still need to put `default_limits[0] if self.vmin is None else self.vmin` in case `self.vmin == 0`, which evaluates as `False` (and same for vmax although I don't think vmax=0 would work very well if at all).

python/lsst/analysis/tools/actions/plot/matrixPlot.py
taranu: Hmm, so you don't want users to be able to turn them off entirely (by supplying an empty list)?

taranu (commented):
A few more notes; I'm assuming the rebase will be trivial.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
enourbakhsh: Good point, done!

python/lsst/analysis/tools/actions/plot/matrixPlot.py
enourbakhsh: Changed it to `if self.xAxisTickValues is not None`; this allows supplying an empty list.

enourbakhsh (commented):
Thank you for the review! I'll proceed with the rebase as soon as you're satisfied with the updates.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -0,0 +1,538 @@
...
+
+        if self.xAxisLabel:
+            ax.set_xlabel(self.xAxisLabel, fontsize=self.axisLabelFontSize)
+
+        if self.yAxisLabel:
+            ax.set_ylabel(self.yAxisLabel, fontsize=self.axisLabelFontSize)
+
+        # Set the colorbar and draw the image.
+        norm = ImageNormalize(vmin=vrange[0], vmax=vrange[1])
+        img = ax.imshow(matrix, interpolation="none", norm=norm)
```
taranu, line 368: I think the `origin` kwarg should be configurable here (I prefer `lower` which is unambiguously the correct choice for our actual CCD images but making it a `ChoiceField[str]` with `upper` as the default is fine if that's the convention for covariance plots for the camera team).

enourbakhsh: Done! I have made it configurable so that we can revisit this decision in the future.

PR #192, opened 2024-01-25
DM-41210: Add tract level injected catalog matching.

jtmccann:


leeskelvin:
For future reference, this PR closed due to our decision to use catalog matching functionality provided by the `MatchProbabilisticTask` in `meas_astrom`.

PR #193, opened 2024-01-25
DM-42655: Add interface for PlotElements

leeskelvin:
Add an interface for basic elements of plots. These are designed to be composed together to create rich plots.


python/lsst/analysis/tools/actions/plot/elements/histElement.py
```diff
@@ -0,0 +1,77 @@
...
+            Keyed data containing the data to plot.
+        ax : `~matplotlib.axes.Axes`
+            Axes to plot on.
+
+        Returns
+        -------
+        data : `~lsst.analysis.tools.interfaces.KeyedData`
+            Data used for plotting.
+        """
+        ax.hist(
```
czwa, line 72: It might be useful to have `cumulative` and `density` as options to pass to `ax.hist`.  Adding them now adds some extra utility to this element that might be useful in the future.

python/lsst/analysis/tools/atools/cpVerifyQuantityProfile.py
```diff
@@ -0,0 +1,155 @@
...
+            15: "C17",
+        }
+
+    def finalize(self):
+        super().finalize()
+
+        # Set the plot element for each panel to an AxPlotElement
+        for key, value in self.produce.plot.valsGroupBy.items():
+            plotElementConfig = PlotElementConfig(
+                plotElement=self.plotElement,
```
czwa, line 135: Am I correct that if I subclassed this, and set `plotElement` based on some arbitrary method, I could mix the `ScatterElement` and `HistElement` in one `GridPlot`?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -0,0 +1,133 @@
...
+
+        for row in range(self.numRows):
+            for col in range(self.numCols):
+                index = row * self.numCols + col
+                if index not in self.valsGroupBy.keys():
+                    continue
+                ax = fig.add_subplot(gs[row, col])
+
+                xList = x.split(",") if (x := self.xDataKeys.get(index)) else None
+                valList = self.valsGroupBy[index].split(",")
```
czwa, line 105: I think this anticipated one of my use cases.  If I specify `valsGroupBy = {0: "C10,C11,C12,C13", 1: "C00,C01,C02,C03"}`, this will fill two subplots with four amplifiers worth of data, correct?

czwa (approved):
One suggestion, and two questions that came from me trying to imagine how to use these classes to solve a number of plotting issues (and wanting to be sure I understand things correctly).

python/lsst/analysis/tools/atools/cpVerifyQuantityProfile.py
leeskelvin: Yes, that's absolutely correct. This is indeed something @natelust and I played around with during testing of `GridPlot`. However, you're also not forced to use `GridPlot` at all - you could equally define your own plot action that calls exactly the plot elements you need for the plot you have in mind.

python/lsst/analysis/tools/actions/plot/elements/histElement.py
leeskelvin: Good idea. I've added `cumulative` and `density`, as well as `color` for good measure and to bring it into line with `scatterElement`.

I wasn't sure what type is appropriate for `cumulative`. I.e., a `bool` captures most use-cases, however, setting `cumulative` to `-1` returns an inverted cumulative histogram (i.e., starting at 1 and counting down). Is that a use-case worth considering? If so, we can change the type to `int` and add extra logic to capture that? Keen to hear your thoughts.

python/lsst/analysis/tools/actions/plot/gridPlot.py
leeskelvin: Indeed you can. Here is an example script that should "just work" for you in a notebook:

```
import lsst.daf.butler as dB
from lsst.analysis.tools.atools.cpVerifyQuantityProfile import (
    CPVerifyQuantityAmpProfileScatterTool,
    CPVerifyQuantityAmpProfileHistTool
)
butler = dB.Butler("/sdf/data/rubin/user/czw/ci_cpp_gen3/DATA")
biasT = butler.get("verifyBiasResults", instrument='LATISS', collections="u/kannawad/ci_cpv_bias")
atool = CPVerifyQuantityAmpProfileScatterTool()
atool.produce.plot.valsGroupBy = {0: "C00,C01,C02,C03", 5: "C10,C11,C12,C13"}
atool.prep.quantityKey = "biasSerialProfile"
atool.finalize()
atool(biasT)
```

As shown, I've modified the first cell to include data from `C00,C01,C02,C03`, and the fifth cell to include data from `C10,C11,C12,C13`. This produces a figure which looks like this:

czwTest

leeskelvin: (Note how the other cells are not drawn, so the `tight_layout` method applied at the end of `gridPlot` has shaved off the superfluous space).

python/lsst/analysis/tools/atools/cpVerifyQuantityProfile.py
leeskelvin: PS - the `plotElementConfig` has been renamed to `gridPanelConfig`, and the grid plot `plotElements` argument has been renamed to `panels`. The over-use of `plotElement` was beginning to get a bit confusing for me, so I thought to broaden our namespace now to more appropriate and descriptive names.

python/lsst/analysis/tools/actions/plot/elements/histElement.py
czwa: I almost always just use `cumulative=True`, so that's enough for me on this ticket.

python/lsst/analysis/tools/atools/cpVerifyQuantityProfile.py
czwa: The new names make sense.  I usually try not to review names unless they're completely confusing, but this seems like a nice improvement.

PR #195, opened 2024-01-26
DM-42685 Add a Tensor interface to analysis tools

natelust:
This commit adds an interface that can be used to represent multi-dimensional data like images, or stacks of images. It uses an industry standard interface to ensure analysis tools can work with a wide variety of data types from numpy arrays to pytorch models.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/_actions.py
+++ b/python/lsst/analysis/tools/interfaces/_actions.py
@@ -45,7 +45,7 @@ import lsst.pex.config as pexConfig
 from lsst.pex.config.configurableActions import ConfigurableAction, ConfigurableActionField
 
 from ..contexts import ContextApplier
-from ._interfaces import KeyedData, KeyedDataSchema, MetricResultType, PlotResultType, Scalar, Vector
+from ._interfaces import KeyedData, KeyedDataSchema, MetricResultType, PlotResultType, Scalar, Tensor, Vector
 
 
 class AnalysisAction(ConfigurableAction):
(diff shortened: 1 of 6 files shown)


python/lsst/analysis/tools/interfaces/_actions.py
```diff
@@ -29,6 +29,7 @@
     "MetricResultType",
     "MetricAction",
     "PlotResultType",
+    "PlotElement",
```
leeskelvin, line 32: Did you intend to add this (and associated class below) on this ticket?

python/lsst/analysis/tools/interfaces/_interfaces.py
```diff
@@ -43,6 +44,36 @@
...
 
+@runtime_checkable
+class Tensor(Protocol):
+    r"""This is an interface only class and is intended to represent data that
+    is 2+ dimensions.
+
+    Technically one could use this for scalars or 1D arrays,
+    but for those the Scalar or Vector interface should be preferred.
+
+    `Tensor`\ s abstract around the idea of a multidimensional array, and work
```
leeskelvin, line 55: What's the purpose of the "\ s"? Is it an attempt to preserve a space? Is it needed? And will it work? I'd imagine you may need to double-escape "\\" here to make this work.

natelust: In the past this has been how you have a link (to Tensor in this case) and have an s at the end of it. You can't include it in the quotes or sphinx can't find the object. the \ is an escape sequence on the space that says don't include this in formatting.

python/lsst/analysis/tools/interfaces/_actions.py
natelust: No, when I switched branches it looks like my editor thought it was code still active in the buffer or something

leeskelvin:
Do you need to add reference to `Tensor` here?

https://github.com/lsst/analysis_tools/blob/6b09da1676c5b9acb6ece641b9103b83fbc4a524/python/lsst/analysis/tools/interfaces/_stages.py#L56

leeskelvin:
PS - see also here:

https://github.com/lsst/analysis_tools/blob/6b09da1676c5b9acb6ece641b9103b83fbc4a524/python/lsst/analysis/tools/actions/keyedData/keyedDataActions.py#L111

PR #196, opened 2024-01-26
DM-42641: Add optional histogram to FocalPlanePlots

cmsaunders:


python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -67,6 +69,19 @@ class FocalPlanePlot(PlotAction):
...
+        optional=True,
+    )
+    plotMax = Field[float](
+        doc="Maximum in z-value to display in the focal plane plot and in the histogram plot, if applicable",
+        default=None,
+        optional=True,
+    )
+    showStats = Field[bool](doc="Show statistics for plotted data", default=True)
+    addHistogram = Field[bool](doc="Add a histogram of all input points", default=False)
+    histBins = Field[int](doc="Number of bins to use in histogram", default=30)
```
taranu, line 84: Consider using `HistPanel` from histPlot rather than having parallel implementations.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -168,8 +183,14 @@ def makePlot(
             noDataFig = addPlotInfo(noDataFig, plotInfo)
             return noDataFig
 
-        fig = plt.figure(dpi=300)
-        ax = fig.add_subplot(111)
+        if self.addHistogram:
+            fig = plt.figure(dpi=300, figsize=(12, 6))
```
taranu, line 187: Any particular reason not to call `plt.subplots(...)` instead?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -168,8 +183,14 @@ def makePlot(
...
+            gs = gridspec.GridSpec(1, 2, width_ratios=[3, 2])
+            ax = fig.add_subplot(gs[0, 0])
+            histAx = fig.add_subplot(gs[0, 1])
+        else:
+            fig = plt.figure(dpi=300)
```
taranu, line 192: We should probably make dpi and figsize configurable some day.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -245,10 +268,26 @@ def makePlot(
         ax.tick_params(labelsize=7)
 
         ax.set_aspect("equal")
+
+        if self.addHistogram:
+            bins = np.linspace(
+                (self.plotMin if (self.plotMin is not None) else min(data["z"].astype(float))),
```
taranu, line 274: If they aren't already, non-finite values should be filtered from `data["z"]` (I didn't see it being done outside of perhaps `sortAllArrays` but that's not done if not `self.showStats`.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -245,10 +268,26 @@ def makePlot(
...
+                (self.plotMax if (self.plotMax is not None) else max(data["z"].astype(float))),
+                self.histBins,
+            )
+            histAx.hist(data["z"].astype(float), bins=bins)
+            histAx.set_xlabel(self.zAxisLabel)
+            histAx.set_ylabel("Bin count")
+            underflow = np.count_nonzero(data["z"] < bins[0])
+            overflow = np.count_nonzero(data["z"] > bins[-1])
+            text = f"Underflow = {underflow}"
+            text += f"\nOverflow = {overflow}"
```
taranu, line 284: Why not just define this all on one line?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -245,10 +268,26 @@ def makePlot(
...
+        if self.addHistogram:
+            bins = np.linspace(
+                (self.plotMin if (self.plotMin is not None) else min(data["z"].astype(float))),
+                (self.plotMax if (self.plotMax is not None) else max(data["z"].astype(float))),
+                self.histBins,
+            )
+            histAx.hist(data["z"].astype(float), bins=bins)
+            histAx.set_xlabel(self.zAxisLabel)
+            histAx.set_ylabel("Bin count")
+            underflow = np.count_nonzero(data["z"] < bins[0])
```
taranu, line 281: Consider adding counts of non-finite and/or nan values if there are likely to be any.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -504,10 +557,26 @@ def makePlot(
         ax.tick_params(labelsize=7)
 
         ax.set_aspect("equal")
+
+        if self.addHistogram:
```
taranu, line 561: I think this function is long enough to justify splitting into single-underscore methods. For example, the 45 lines from `divider = make_axes_locatable(ax)` to `fig = addPlotInfo(fig, plotInfo)` are identical except for the arg passed to `fig.colorbar`.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: If I just use `HistPanel` for the configuration, there are a lot of additional config options I would need to implement. Or I could fully take advantage of the `HistPlot` class and make an instance of `HistPlot`, then use `HistPlot._make_panel()`, but obviously this is private. Do you have something else in mind?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: I would generally use `plt.subplots()`, but @mfisherlevine and I pair-coded this, and he has strong negative feelings about using `plt.<anything>` other than `plt.figure()`. After the fact, I think we determined that `plt.subplots()` is effectively no worse than `plt.figure()`, so I may go back to this.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: Agreed

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: I opted to use the `...math.nanMin()` and `...math.nanMax()` methods.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
cmsaunders: Added "Non-finite = ..."

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
taranu: I haven't looked at the generic multi-panel plot ticket yet, but in the meantime I think you can add a config field like `hist = ConfigurableActionField[HistPlot]`.

Looking at `HistPanel` a bit more closely, I think `HistPlot._make_panel` should essentially move wholesale to `HistConfig.make_panel`; the only place `self` is needed is to refer to `self.panels[panel]` and to other single-underscore methods that similarly could be moved to `HistConfig`. But that is a non-trivial refactor that probably should go on another ticket.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
taranu: Ok, I thought subplots just conveniently made more or less the same calls as what you had there anyway.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
taranu: I don't have an issue with calling `_make_panel` - that should be recommended usage (but like I said, it's not a trivial fix, it needs to be split off from `HistPlot` and renamed `make_panel`).

PR #197, opened 2024-01-27
Tickets/DM-42648: Create FocalPlaneGeometryPlot for Readnoise median

jrmullaney:
The changes shown here are actually from three related tickets (DM-42671, DM-42651, DM-42648), so there are three different branches. The branches are, however, linear, so they are effectively a set of three commits. This branch (DM-42648) is the last of the branches, so by merging this branch, the changes made in all branches will be merged into main.

### DIFF
--- a/python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
+++ b/python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
@@ -278,6 +278,15 @@ class FocalPlaneGeometryPlot(FocalPlanePlot):
         },
     )
 
+    def getInputSchema(self, **kwargs) -> KeyedDataSchema:
+        base = []
+        base.append(("detector", Vector))
+        if self.level == "amplifier":
+            base.append(("amplifier", Vector))
+        base.append(("z", Vector))
+
+        return base
+
     def makePlot(
         self,
         data: KeyedData,
--- a/python/lsst/analysis/tools/atools/calibration.py
+++ b/python/lsst/analysis/tools/atools/calibration.py
@@ -1,0 +1,52 @@
+# This file is part of analysis_tools.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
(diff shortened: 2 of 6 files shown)


python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -278,6 +278,15 @@ class FocalPlaneGeometryPlot(FocalPlanePlot):
         },
     )
 
+    def getInputSchema(self, **kwargs) -> KeyedDataSchema:
+        base = []
+        base.append(("detector", Vector))
+        if self.level == "amplifier":
```
sr525, line 284: Can level ever be anything else? What is the default? Where is it defined? If we only forsee it being amplifier or nothing then maybe a True/False config option would be better.

python/lsst/analysis/tools/atools/calibration.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.process.buildActions.z = LoadVector()
+        self.process.buildActions.z.vectorKey = "biasReadNoise"
+        self.process.buildActions.detector = LoadVector()
+        self.process.buildActions.detector.vectorKey = "detector"
+        self.process.buildActions.amplifier = LoadVector()
+        self.process.buildActions.amplifier.vectorKey = "amplifier"
+
+        self.produce.plot = FocalPlaneGeometryPlot()
+        # Note: In anticipation of the addHistogram option
```
sr525, line 47: should be TO DO: rather than note.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
jrmullaney: `level` can be either "amplifier" or "detector". The default is "amplifier", but if it is set to "detector", then the user shouldn't need to load the "amplifier" vectorKey. It's defined on line 272 in the same file.

sr525: Sorry, for some reason I thought this was a new option added on this ticket. Opps.

PR #203, opened 2024-02-01
DM-42741: Make focalPlaneGeometryPlot look a bit better

sr525:


python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -343,6 +355,10 @@ def makePlot(
         fig : `matplotlib.figure.Figure`
             The resulting figure.
         """
+
+        cmap = mkColormap(["midnightBlue", "lightcyan", "darkgreen"])
```
plazas, line 359: Why are there 3 options here, and could we add "tableau-colorblind10"?

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
```diff
@@ -487,8 +503,23 @@ def makePlot(
...
+            vmin = statMed - 4.0 * statMad
+        if self.vmax:
+            vmax = self.vmax
+        else:
+            vmax = statMed + 4.0 * statMad
+
+        valuesPlot = np.clip(values, vmin, vmax)
+
+        patchCollection = PatchCollection(
+            patches, edgecolor="white", cmap=cmap, linewidth=0.5, linestyle=(0, (0.5, 3))
```
plazas, line 520: would it be too much to make these options (e.g., edgecolor, linewidth...) configurable too, or is it not worth it?

plazas (approved):
LGTM! Just a few questions for clarification.

python/lsst/analysis/tools/actions/plot/focalPlanePlot.py
sr525: I would prefer not to make all of these things configurable. I would like the plots produced to be uniform and not clutter up the code with too many config options.

PR #204, opened 2024-02-08
DM-42820: Add method to load an AnalysisTool from a pipeline

natelust:
Add a class method to AnalysisTool which allows a tool defined in a pipeline to be loaded into a python session.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/_analysisTools.py
+++ b/python/lsst/analysis/tools/interfaces/_analysisTools.py
@@ -25,11 +25,14 @@ __all__ = ("AnalysisTool",)
 
 from collections.abc import Mapping
 from functools import wraps
+from operator import attrgetter
 from typing import Callable, Iterable, Protocol, runtime_checkable
 
 import lsst.pex.config as pexConfig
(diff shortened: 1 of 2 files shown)


python/lsst/analysis/tools/interfaces/_analysisTools.py
```diff
@@ -282,6 +284,51 @@ def getOutputNames(self, config: pexConfig.Config | None = None) -> Iterable[str
...
+        """Construct an `AnalysisTool` from a definition written in a
+        `~lsst.pipe.base.Pipeline`.
+
+        Parameters
+        ----------
+        pipeline : `str` or `~lsst.pipe.base.Pipeline`
+            The pipeline to load the `AnalysisTool` from.
+        name : `str`
+            The name of the analysis tool to run. This can either be just the
+            name assigned to the tool, or an absolute name a config hierarchy
```
leeskelvin, line 298: "or an absolute name _of_ a config hierarchy"?

leeskelvin: or _in_ a...

PR #207, opened 2024-02-14
DM-42906: Edit scalarActions to accommodate Tensors

jrmullaney:
Not all these changes are strictly necessary to address the ticket, but I thought it better have consistency across actions where possible.

### DIFF
--- a/python/lsst/analysis/tools/actions/scalar/scalarActions.py
+++ b/python/lsst/analysis/tools/actions/scalar/scalarActions.py
@@ -205,7 +205,7 @@ class ApproxFloor(ScalarFromVectorAction):
 
     def __call__(self, data: KeyedData, **kwargs) -> Scalar:
         mask = self.getMask(**kwargs)
-        values = np.sort(_dataToArray(data[self.vectorKey.format(**kwargs)])[mask])  # type: ignore
+        values = np.sort(_dataToArray(data[self.vectorKey.format(**kwargs)])[mask], axis=None)  # type: ignore
         x = values.size // 10
         return nanMedian(values[-x:])
 
--- a/python/lsst/analysis/tools/actions/scalar/scalarActions.py
+++ b/python/lsst/analysis/tools/actions/scalar/scalarActions.py
@@ -57,6 +57,17 @@ from ...math import nanMax, nanMean, nanMedian, nanMin, nanSigmaMad, nanStd
 log = logging.getLogger(__name__)
 
 
+def _dataToArray(data):
+    """Convert input data into a numpy array using the appropriate
+    protocol. `np.from_dlpack` is used for Tensor-like arrays
+    where possible.
+    """
+    try:
+        return np.from_dlpack(data)
+    except AttributeError:
+        return np.array(data)
+
+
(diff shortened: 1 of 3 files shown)


python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -67,8 +67,8 @@ class MedianAction(ScalarFromVectorAction):
 
     def __call__(self, data: KeyedData, **kwargs) -> Scalar:
         mask = self.getMask(**kwargs)
-        values = data[self.vectorKey.format(**kwargs)][mask]
-        med = nanMedian(values) if len(values) else np.NaN
+        values = np.array(data[self.vectorKey.format(**kwargs)])[mask]
```
natelust, line 70: Ok, so this is going to seem a bit weird, as all worrying about typing is in python. As of now Vector and Tensor are two different things, and actually have two different interfaces. As such we really should be checking for each. For instance:
```
values = np.from_dlpack(info) if isinstance((info := data[self.vectorKey.format(**kwargs)]), Tensor) else np.array(info)
values = values[mask]
```
The reason to do this is things that satisfy the Tensor protocol can be turned into numpy arrays with zero copies, but not everything that satisfies the Vector interface satisfy the Tensor interface, such as pandas sequences. (In this case it turns out they can be turned into arrays with zero copy, but as a quirk with an old style numpy interface).
We can of course support more restrictive interfaces, but up to this point we have been trying hard to support whatever data someone might pass in.

jrmullaney: Thanks. I've made the suggested change by adding a `_dataToArray` function at the start of the file, which does as you suggest and is called throughout. I felt it better to do that rather than replicate the same code multiple times throughout. Since the output of the function will always be a np.array, I don't _think_ we need to wrap them with `cast(Vector, ...)` any more, but I'm still learning about typing.

sr525:
Have you run the core pipelines to check that they all still work?

jrmullaney:
> Have you run the core pipelines to check that they all still work?

Not yet. I wasn't sure if I was 100% on the right track, so didn't want to run jenkins until toward the end the process. I also need to rebase, so I'd do it after that.

jrmullaney:
Hi @natelust @sr525
I'd like to reawaken this from hibernation. As you'll see, I've edited `_dataToArray` at the top of `scalarActions.py` so that it only attempts to `from_dlpack` float or integer arrays, since those are the only array types that `from_dlpack` can cope with.
I've checked, and none of the core or extended pipelines break in response to this change.

tests/test_actions.py
```diff
@@ -78,17 +78,42 @@ def setUp(self):
         x[41] = np.nan
         x[59] = np.nan
         y = x**2
-        self.data = {"r_y": x, "i_y": y}
-        self.mask = np.zeros(100, dtype=bool)
-        self.mask[1:50] = 1
+        self.data = {
+            "g_y": pd.Series(y),
```
sr525, line 82: No pandas! Please remove all pandas specific stuff.

sr525 (approved):

PR #209, opened 2024-02-18
DM-42941: Using default axis labels for astrom and photom focalPlanePlots

jrmullaney:
Default axis labels include units (mm).


cmsaunders (approved):
This looks good. You probably want to change these lines too: https://github.com/lsst/analysis_tools/blob/14baeae61413179daf22febd227556001f925e49/python/lsst/analysis/tools/atools/calibration.py#L54

PR #210, opened 2024-02-20
DM-42888: Incorporate amp-to-amp correlation plots in analysis_tools for camera diagnostics

enourbakhsh:


python/lsst/analysis/tools/actions/plot/matrixPlot.py
```diff
@@ -403,10 +433,10 @@ def makePlot(self, data: KeyedData, plotInfo: Mapping[str, str] | None = None, *
         shift = 0.5 if self.setPositionsAtPixelBoundaries else 0
 
         if self.component1Key is not None and self.component2Key is not None:
-            xAxisTickValues = np.arange(matrix.shape[0] + shift)
-            yAxisTickValues = np.arange(matrix.shape[1] + shift)
-            xAxisTickLabels = {key + shift: str(val) for key, val in zip(range(matrix.shape[0]), comp1[0, :])}
-            yAxisTickLabels = {key + shift: str(val) for key, val in zip(range(matrix.shape[1]), comp2[:, 0])}
+            xAxisTickValues = np.arange(matrix.shape[1] + shift)
```
czwa, line 436: Can you explain why you flipped indices here?  I don't think it matters, but I was curious.

czwa (approved):
One minor question for my own understanding.

python/lsst/analysis/tools/actions/plot/matrixPlot.py
enourbakhsh: Yeah, it doesn't quite matter here since we're dealing with square matrices for correlations. Just thought it might be useful for future-proofing, in case someone comes along with a non-square matrix to plot. I thought the flip from `matrix.shape[0]` to `matrix.shape[1]` for the x-axis made sense, as in a 2D matrix (array), `matrix.shape[0]` represents the number of rows (which can be thought of as the matrix's height?), and `matrix.shape[1]` represents the number of columns (which can be thought of as the matrix's width?), unless I'm totally mistaken.

czwa: That all makes sense to me.  Thanks for the quick explanation!

PR #212, opened 2024-02-23
DM-42157: Skip scatterplot bin plot for non-finite sigMadY

taranu:
This came up with some MultiProFit plots on ci_imsim. Add it to the list of easy fixes that ought not to be necessary if all of the relevant flags/selection were applied to the input data, I suppose.


fred3m (approved):
LGTM

PR #216, opened 2024-03-08
DM-43238: Refactor calcDistances.py for a 50-100x speed increase in AnalyzeMatchedVisitCore

erykoff:
This refactors the loops in calcDistances to be vectorized and get a 10x speed increase.  In addition, it adds downsampling which gets another 5-10x speed increase in the densest fields.

One algorithmic change from the old version is that now the computation of `sepResiduals` changes from from `separations - median(separations)` to `separations - mean(separations)` because this is much easier to vectorize.

There is also a small fix for empty tracts.

### DIFF
--- a/python/lsst/analysis/tools/actions/keyedData/calcDistances.py
+++ b/python/lsst/analysis/tools/actions/keyedData/calcDistances.py
@@ -106,10 +106,8 @@ class CalcRelativeDistances(KeyedDataAction):
             h, rev = esutil.stat.histogram(arrayIn, rev=True)
             arrayOut = np.zeros(len(arrayIn), dtype=np.int32)
             (good,) = np.where(h > 0)
-            counter = 0
-            for ind in good:
+            for counter, ind in enumerate(good):
                 arrayOut[rev[rev[ind] : rev[ind + 1]]] = counter
-                counter += 1
             return arrayOut
 
         groupId = _compressArray(data[self.groupKey])
@@ -223,11 +221,12 @@ class CalcRelativeDistances(KeyedDataAction):
         sepResiduals = separations - sepMean[matchedPairInd]
         sepResiduals = sepResiduals[np.isfinite(sepResiduals)]
 
-        if len(rmsDistances) == 0:
-            AMx = np.nan * u.marcsec
-        else:
-            AMx = (np.median(rmsDistances) * u.radian).to(u.marcsec)
+        # This is always going to be valid because we checked the number
+        # of good pairs above.
+        AMx = (np.median(rmsDistances) * u.radian).to(u.marcsec)
 
+        # Because there is a more stringent selection for sepResiduals,
+        # we need to check that we have enough to compute the metrics.
         if len(sepResiduals) <= 1:
             AFx = np.nan * u.percent
             ADx = np.nan * u.marcsec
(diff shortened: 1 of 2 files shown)


python/lsst/analysis/tools/actions/keyedData/calcDistances.py
```diff
@@ -81,61 +89,139 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
...
+        # Match together pairs that have the same visit.
+        # It is unfortunate that this requires a loop, but it is not slow.
+        # After this matching we have a set of matchedObsInd1/matchedObsInd2
+        # that are all individual observations that are in the annulus and
+        # share a visit. The matchedPairInd groups all the paired observations
+        # of a given pair.
+        matchedObsInd1 = []
+        matchedObsInd2 = []
+        matchedPairInd = []
+        for ind in range(len(i1)):
```
cmsaunders, line 175: I didn't test this, but I think you could do this loop over visits instead of over matches, which would mean fewer iterations of the loop. Since you say it's not slow though, it's probably not worth reworking.

python/lsst/analysis/tools/actions/keyedData/calcDistances.py
```diff
@@ -81,61 +89,139 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
...
+        sepStd[good] = np.sqrt(sepStd[good] / (nSep[good] - 1))
+        rmsDistances = sepStd[good]
+
+        # Need sepResiduals, but only when nSep is > 2.
+        bad2 = nSep <= 2
+        sepMean[bad2] = np.nan
+        sepResiduals = separations - sepMean[matchedPairInd]
+        sepResiduals = sepResiduals[np.isfinite(sepResiduals)]
 
         if len(rmsDistances) == 0:
```
cmsaunders, line 226: I don't think you need this anymore, since you already returned if `good.sum() == 0`

cmsaunders (approved):

python/lsst/analysis/tools/actions/keyedData/calcDistances.py
```diff
@@ -81,61 +89,139 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
...
+        # Check if tract is overlapping ra=0 and rotate if so.
+        # We assume a tract is smaller than 60x60 degrees.
+        rotation = 0.0
+        if np.max(data[self.raKey]) > 330.0 and np.min(data[self.raKey]) < 30.0:
+            rotation = 180.0
+            raRotated = np.array(data[self.raKey]) - rotation
+        else:
+            raRotated = np.array(data[self.raKey])
+
+        np.add.at(meanRa, groupId, raRotated)
```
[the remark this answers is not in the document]
erykoff: Numpy universal functions `at` are my favorite thing, and they even made them faster in the version that will be in the next rubin-env.

python/lsst/analysis/tools/actions/keyedData/calcDistances.py
erykoff: You know, I thought that too, but couldn't figure out how to make it work with the matching necessary.  Although I do think an even more optimized version (maybe even without this loop) is possible, I had to just stop and accept the speed increase I had gotten without making it perfect.

PR #218, opened 2024-03-08
DM-41108: Temporarily remove broken AP metrics

isullivan:
To be added back in DM-43201

### DIFF
--- a/pipelines/apDetectorVisitQualityCore.yaml
+++ b/pipelines/apDetectorVisitQualityCore.yaml
@@ -1,6 +1,11 @@
 description: |
   AP plots and metrics to assess CcdVisit quality in Prompt Processing
   Additional AP metrics may be written in apDetectorVisitQualityExtended
+parameters:
+  # Note that these parameter names are shared with ApPipe, and are overwritten
+  # by changes in the values there.
+  coaddName: goodSeeing
+  fakesType: ''
 tasks:
   analyzeAssocDiaSrcCore:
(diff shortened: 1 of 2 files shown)


pipelines/apDetectorVisitQualityCore.yaml
kfindeisen, line 1: Neither `apDetectorVisitQualityCore` nor "Tier 1 AP plots and metrics to assess CcdVisit quality" really makes it clear how these metrics will be used.

Is the plan that this file can only contain metrics that wouldn't be disruptive to real-time processing? If so, a warning about the restriction should be present in the name, description, or both. (I'm assuming, of course, that there'd be a bigger and slower set of metrics somewhere for batch or daytime processing...)

kfindeisen (approved):
No concerns about the change itself, but I'd like to see some safeguards to make sure these metrics stay fast and slim.

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -1,6 +1,9 @@
 description: |
   AP plots and metrics to assess CcdVisit quality in Prompt Processing
   Additional AP metrics may be written in apDetectorVisitQualityExtended
+parameters:
+  coaddName: goodSeeing
+  fakesType: ''
```
kfindeisen, line 6: Does this need a comment saying that the parameter names are shared with `ApPipe` (I assume that's the intent)?

kfindeisen (approved):
Looks good, thanks for the clean commit history!

PR #219, opened 2024-03-13
DM-42088 Add a check for invalid characters in measurement names

natelust:


python/lsst/analysis/tools/atools/photometry.py
```diff
@@ -86,9 +86,9 @@ def setDefaults(self):
         self.produce.metric.units = {"median": "mmag", "sigmaMad": "mmag", "mean": "mmag"}
 
         self.produce.metric.newNames = {
-            "median": "{band}_ap12-psf_median",
-            "mean": "{band}_ap12-psf_mean",
-            "sigmaMad": "{band}_ap12-psf_sigmaMad",
+            "median": "{band}_ap12_diff_psf_median",
+            "mean": "{band}_ap12_diff_psf_mean",
+            "sigmaMad": "{band}_ap12_diff_psf_sigmaMad",
```
leeskelvin, line 91: If we're keeping this in-line with the prior renaming exercise on DM-42062, these should be:
```
            "median": "{band}_ap12_psf_diff_median",
            "mean": "{band}_ap12_psf_diff_mean",
            "sigmaMad": "{band}_ap12_psf_diff_sigmaMad",
```
I.e., the Ap12 and PSF difference.

python/lsst/analysis/tools/interfaces/_stages.py
```diff
@@ -182,13 +185,23 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
         return results
 
 
+def _newNameChecker(value: str) -> bool:
+    if "-" in value:
+        # Yes this should be a log here, as pex config give us no other
+        # useful way to get info into the exception.
```
leeskelvin, line 191: Rather than a raise?

Typo: `give` to `gives`.

leeskelvin (approved):
Looks good. One comment on naming things, and one on a raise clarification.

python/lsst/analysis/tools/interfaces/_stages.py
natelust: Yeah, because I want the pex_execution framework to raise a `FieldValidationError` which it will do with the check function. That exception is the one that various tools will be looking for, but there also not an easy way to raise it yourself (you need access to various objects) I thought about hacking various complicated things to make it do what you might expect, but I figured it was not worth the time. This will log an exact message, and the standard exception WILL have  a message containing the actual text of the invalid value.

PR #221, opened 2024-03-17
DM-43231: Create Analysis Tool to send visitSummary info to sasquatch

jrmullaney:
This ticket involves changes to two repositories: this repository (analysis_tools) and pipe_tasks. Here, I'm adding a new analysis tool task and atool to extract summary statistics from a calexp's metadata and write them to sasquatch.

### DIFF
--- a/python/lsst/analysis/tools/atools/calexpMetrics.py
+++ b/python/lsst/analysis/tools/atools/calexpMetrics.py
@@ -1,0 +1,73 @@
+# This file is part of analysis_tools.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("CalexpSummaryMetrics",)
+
+from ..interfaces import AnalysisTool
+
+
+class CalexpSummaryMetrics(AnalysisTool):
+    """
+    Class to load statistics from the summary stats contained with a calexp's
+    metadata and write them to metrics.
+    """
+
+    propagateData: bool = True
+
+    # raCorners and decCorners statistics cannot be written to a metric,
+    # as metrics can only be single-valued (i.e., scalars).
+    _units = {
+        "psfSigma": "pixel",
+        "psfArea": "",  # There is no astropy sq. pixel unit.
+        "psfIxx": "pixel",
+        "psfIyy": "pixel",
+        "psfIxy": "pixel",
+        "ra": "degree",
+        "dec": "degree",
+        "zenithDistance": "degree",
+        "zeroPoint": "mag",
+        "skyBg": "mag",
+        "skyNoise": "mag",
+        "meanVar": "mag",
+        "astromOffsetMean": "arcsec",
+        "astromOffsetStd": "arcsec",
+        "nPsfStar": "",
+        "psfStarDeltaE1Median": "pixel",
+        "psfStarDeltaE2Median": "pixel",
+        "psfStarDeltaE1Scatter": "pixel",
+        "psfStarDeltaE2Scatter": "pixel",
+        "psfStarDeltaSizeMedian": "pixel",
+        "psfStarDeltaSizeScatter": "pixel",
+        "psfStarScaledDeltaSizeScatter": "pixel",
+        "psfTraceRadiusDelta": "pixel",
+        "maxDistToNearestPsf": "deg",
+        "effTime": "s",
+        "effTimePsfSigmaScale": "s",
+        "effTimeSkyBgScale": "s",
+        "effTimeZeroPointScale": "s",
+    }
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.prep.keysToLoad = list(self._units.keys())
+        self.produce.metric.units = self._units
--- a/python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
+++ b/python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
@@ -25,16 +25,16 @@ __all__ = (
     "CalexpSummaryAnalysisTask",
 )
 
+from lsst.pipe.base import InputQuantizedConnection, OutputQuantizedConnection, QuantumContext
 from lsst.pipe.base import connectionTypes as cT
 
-from lsst.pipe.base import InputQuantizedConnection, OutputQuantizedConnection, QuantumContext
 from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
 
 
 class CalexpSummaryAnalysisConnections(
     AnalysisBaseConnections,
     dimensions=("visit", "band", "detector"),
-    defaultTemplates={"inputName": "calexp", "outputName": "calexpSummary"},
+    defaultTemplates={"inputName": "calexp.summaryStats", "outputName": "calexpSummary"},
 ):
     data = cT.Input(
         doc="Calibrated exposure summary statistics to load from the butler",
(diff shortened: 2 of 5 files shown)


ctslater (commented):
One small comment.

pipelines/calexpQualityCore.yaml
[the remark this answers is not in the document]
jrmullaney: I'm inclined to disagree: I feel that there is a need. This is in case someone wishes to create the metrics using a standalone analysis tool (i.e., the usual way we use analysis tools). For example, something gets flagged by the Step1 running, so then wants to reproduce the "weirdness" without running the calibrate step.

I did think about putting this in visitQualityCore, but that seems to aggregate/plot data for a whole visit (i.e., across all detectors), whereas this is for a single detector.

Having said all that, this doesn't affect its use for OR3, so I can drop it if needed and we can come back to it later.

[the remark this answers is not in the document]
natelust: I too am a little bit concerned that this pipeline hanging out there will confuse people as to why it's not included and run. At minimum I think that means it needs a lot of documentation added to it. I don't really think it is needed and can be deleted. It is no real extra work to do `pipetask run -b ... -p $ANALYSIS_TOOLS_DIR/pipelines/calexpQualityCore.yaml ...` vs `pipetask run -b ... -t last.analysis.tools.tasks.CalexpSummaryAnalysisTask ...` if someone wants to run a one off to regenerate something for investigation. And if they are doing investigation, it might make sense for them to use it through the terminal/notebook interface anyway using the tool directly.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -39,6 +39,7 @@
     "MedianHistAction",
     "IqrHistAction",
     "DivideScalar",
+    "LoadScalar",
```
natelust, line 42: In context of the commit I pushed (if there is no issues with it you see in your cross review) I think this should just be deleted.

python/lsst/analysis/tools/atools/calexpMetrics.py
```diff
@@ -0,0 +1,74 @@
...
+        "effTime": "s",
+        "effTimePsfSigmaScale": "s",
+        "effTimeSkyBgScale": "s",
+        "effTimeZeroPointScale": "s",
+    }
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        for key in self._units.keys():
```
natelust, line 71: in context of the commit I added this just becomes `self.prep.keysToLoad = list(self._units.keys)`

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
```diff
@@ -0,0 +1,69 @@
...
+from lsst.pipe.base import connectionTypes as cT
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class CalexpSummaryAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions=("visit", "band", "detector"),
+    defaultTemplates={"inputName": "calexp", "outputName": "calexpSummary"},
+):
```
natelust, line 38: I am not sure this is what you want. In the calibrate task I think you can just write out the `ExposureSummaryStats` in a connection, and then just read that in here, that would be way faster and cheaper than loading the whole exposure. There is already a Storage Class type for it in the butler: https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/configs/storageClasses.yaml#L33C3-L33C23 . This will need a change on the pipe_base side of the ticket.

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
```diff
@@ -0,0 +1,69 @@
...
+        self,
+        butlerQC: QuantumContext,
+        inputRefs: InputQuantizedConnection,
+        outputRefs: OutputQuantizedConnection,
+    ) -> None:
+        # Docstring inherited.
+
+        inputs = butlerQC.get(inputRefs)
+
+        summary = inputs["data"].getInfo().getSummaryStats().__dict__
```
natelust, line 66: if you change the input connection you will only need the `__dict__` here

natelust (approved):

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
natelust: BUT if we don't want to write the summary stats out for some good reason, this here looks fine.

pipelines/calexpQualityCore.yaml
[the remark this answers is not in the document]
sr525: I'm going to be adding some pipelines soon that won't be run as standard. We need some not run in processing but ready to go if something goes wrong pipelines. For example to be triggered if metrics cross a threshold. We can improve the docs and maybe make a readme in the pipelines directory?

python/lsst/analysis/tools/atools/calexpMetrics.py
jrmullaney: Is it ok to use `self.metric_tags = list(self._units.keys)`? It seems to work fine, but there may be a reason I'm not aware of for not using it.

pipelines/calexpQualityCore.yaml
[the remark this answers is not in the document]
jrmullaney: Can we leave it for now? I'd like to get this merged by tomorrow, and this isn't critical for that.

python/lsst/analysis/tools/atools/calexpMetrics.py
natelust: oh no, that should be `self._units.keys()`. You want to iterate though the dict keys, I just missed typing the (). If it works without them I think something is up, and I would verify all the code produces the output you think.

pipelines/calexpQualityCore.yaml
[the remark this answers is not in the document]
natelust: if you want to get this merged fast, I would say delete this from this ticket, but don't delete it completely and talk with Sophie about longer term plans after this ticket. The problem is once it exists you have to worry about legacy users of this as an interface if you want to move or drop it later. Pipetask _dose_ have the `-t` interface so it is not like someone can't run it stand alone, and I don't like the idea of a pipeline existing and named so specifically as this just to help someone not type out the class on the command line. I do support what Sophie is talking about as they are intended for specific types of followups (i.e. photometry) not so specific as pipelines for one task.

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
natelust: I am fine for time reasons if you want to skip this for this ticket

jrmullaney: I can't see where in calibrate.py `ExposureSummaryStats` are being written, so while I agree it would be much cheaper to just load the stats, it doesn't seem to be possible at thiss stage. Maybe I've missed something, though.

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
jrmullaney: See above.

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
natelust: Actually, is there a reason to load it like this at all for this? It make sense if you wanted to stream them out as you go, but if you are all the way in analysis tools running a pipeline for follow up, why would you not just get it from the summary table that has the whole visit in it?

python/lsst/analysis/tools/atools/calexpMetrics.py
jrmullaney: Sorry, yes, I also meant `self._units.keys()` (with the brackets).

python/lsst/analysis/tools/tasks/calexpSummaryAnalysis.py
jrmullaney: I agree that one may want to just do as you suggest. However, I'm looking at it as someone who isn't very familiar with analysis_tools - if calexpSimmaryAnalysis is retargeted in the yaml file, then it's immediately clear to someone that they can use that analysis tool task to replicate the metrics.

natelust: Can you try changing your connection to `calexp.summaryStats` talking to Eli I think this is not possible with component disassembly.

jrmullaney: calexp.summaryStats works, so I can change it to that (if that's what you mean).

jrmullaney: I've changed it to `calexp.summaryStats`.

pipelines/calexpQualityCore.yaml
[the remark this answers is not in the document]
sr525: The core in the name implies it should be run every time. Let's leave it out and then have a chat about future pipelines. I could use a rubber duck if nothing else.

PR #223, opened 2024-03-18
DM-42823: add metrics associated with StellarPhotometricResidualsFocalPlane in analysis_tools

jeffcarlin:


jeffcarlin:
Jenkins run kicked off at [this link](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/902/pipeline)

python/lsst/analysis/tools/atools/photometricRepeatability.py
```diff
@@ -193,5 +199,21 @@ def setDefaults(self):
...
+            "photResidFocalPlaneSigmaMad": "mmag",
+            "photResidFocalPlaneStdev": "mmag",
+            "photResidFocalPlaneMedian": "mmag",
+        }
+
+        self.produce.metric.newNames = {
+            "photResidFocalPlaneSigmaMad": "{band}_photResidFocalPlaneSigmaMad",
+            "photResidFocalPlaneStdev": "{band}_photResidFocalPlaneStdev",
+            "photResidFocalPlaneMedian": "{band}_photResidFocalPlaneMedian",
+        }
```
cmsaunders, line 219: These quantities don't really have any connection with the focal plane, so I don't think you want to have "FocalPlane" in the metric name. Maybe "Tract" instead, since this is calculated on all the sources in a tract?

cmsaunders (commented):
This looks good, though I think `photResidFocalPlaneMedian` is going to be zero by definition, since you are calculating it on quantities with the median subtracted. I am also wondering how to think about `photResidFocalPlaneStdev` as compared to `photRepeatStdev` calculated in `StellarPhotometricRepeatability` above. 
I have one comment about the naming, based on the fact that the focal plane position doesn't actually come into the calculations.

jeffcarlin:
> This looks good, though I think `photResidFocalPlaneMedian` is going to be zero by definition, since you are calculating it on quantities with the median subtracted. 

I thought about this, too. I think that since the _mean_ is subtracted off, there may be cases where the median of the residuals is not zero. But I'm not positive about that, and would be fine with just removing the median.

> I am also wondering how to think about `photResidFocalPlaneStdev` as compared to `photRepeatStdev` calculated in `StellarPhotometricRepeatability` above. 

It would be interesting to compare -- the value from `StellarPhotometricRepeatability` is the internal repeatability, while the photResid version is the scatter about the reference value. I feel like those shouldn't necessarily always be the same, but I'm not sure...

python/lsst/analysis/tools/atools/photometricRepeatability.py
jeffcarlin: Good point -- only the plotting is actually happening over the focal plane, while the metric is over a tract. I'll change them to "Tract" instead.

cmsaunders:
But I think it is the median that is subtracted off (https://github.com/lsst/analysis_tools/blob/c0371565ee86fed08fda65a3ba1371436423286a/python/lsst/analysis/tools/atools/photometricRepeatability.py#L191). Looking at this again, I realized that the `zAxisLabel` is actually not correct then:         ```self.produce.plot.zAxisLabel = "Mag - Mag$_{mean}$ (mmag)"``` should have "Mag$_{median$". I guess it makes sense to leave this in though, because when you add the detector-level and visit-level calculations, the median will be a meaningful quantity.
I agree the two standard deviation metrics won't be the same. I guess the issue is making sure that they have distinct enough names, but using "Tract" for the new one helps.

jeffcarlin:
For (re-)naming, how about "{band}_photResidVsRefcatTractMedian", etc.?

Good catch on the median vs. mean -- I can fix the plot label on this ticket, too.

cmsaunders:
I don't think this task uses the reference catalogs at all--the residual is between the source magnitudes and the median for all the sources corresponding to the same match.

jeffcarlin:
> I don't think this task uses the reference catalogs at all--the residual is between the source magnitudes and the median for all the sources corresponding to the same match.

Oops, right. I was conflating two things I was thinking about. I changed the names to photResidTractSigmaMad, photResidTractStdev, and photResidTractMedian.

cmsaunders:
Sounds good to me.

PR #228, opened 2024-03-21
DM-43431: In SasquatchDatastore allow namespace and extra fields to come from environment

timj:


python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -101,7 +114,17 @@ def __init__(
 
         self.accessToken = self.config.get("accessToken", "na")
 
-        self.namespace = self.config.get("namespace", "lsst.dm")
+        default_namespace = os.environ.get("DAF_BUTLER_SASQUATCH_NAMESPACE", "lsst.dm")
+        self.namespace = self.config.get("namespace", default_namespace)
```
ktlim, line 118: Isn't this overriding backwards?  I thought we want `self.config["namespace"]` unless `DAF_BUTLER_SASQUATCH_NAMESPACE` is defined, in which case we take it.

timj: Hmm. Right. Environment variable should always override. I will redo this.

ktlim: I think the same for the extras, for consistency, even if that one might go either way.

timj: Yes. Fixed.

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -80,6 +81,9 @@ class SasquatchDatastore(GenericBaseDatastore):
     namespace: str
     """The namespace in Sasquatch where the uploaded metrics will be
     dispatched.
+
+    Defaults to the value obtained from the ``$DAF_BUTLER_SASQUATCH_NAMESPACE``
+    environment variable and "lsst.dm" if no environment variable is defined.
```
kfindeisen, line 86: The wording here is a bit confusing, assuming I understand correctly that these fields can _all_ be set directly in config. Maybe:
```suggestion
    Overridden by the ``$DAF_BUTLER_SASQUATCH_NAMESPACE`` environment variable
    if it is defined. Defaults to "lsst.dm".
```

kfindeisen: Should this have an explicit default like `defaultConfigFile` does?

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -86,6 +86,16 @@ class SasquatchDatastore(GenericBaseDatastore):
...
     """
 
+    extra_fields: dict[str, str | int | float] | None
+    """Extra key/value pairs that should be passed along with the metric
+    when storing in Sasquatch.
+
+    If not given extra key/value pairs can be obtained from the
+    ``$SASQUATCH_EXTRAS`` environment variable. This environment variable
+    must have the form of "k1=v2;k2=v2". If no environment variable is given
+    defaults can be read from the datastore configuration.
```
kfindeisen, line 96: Same comment about understanding what trumps what. Maybe:
```suggestion
    Overridden by the ``$SASQUATCH_EXTRAS`` environment variable if it is
    defined; this environment variable must have the form of "k1=v2;k2=v2".
    The key-value pairs in ``$SASQUATCH_EXTRAS`` and the datastore
    configuration are *NOT* merged.
```
(Though given my comment below, maybe merging was the intended behavior?)

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -110,6 +120,16 @@ def __init__(
...
 
+        extra_fields: dict[str, str | int | float] | None = None
+        if extras_str := os.environ.get("SASQUATCH_EXTRAS"):
+            extra_fields = {}
+            for item in extras_str.split(";"):
+                k, v = item.split("=")
+                extra_fields[k] = v
+            self.extra_fields = extra_fields
+        else:
+            self.extra_fields = self.config.get("extra_fields", extra_fields)
```
kfindeisen, line 131: On this branch, `extra_fields` is always `None`... was that intentional?

kfindeisen (approved):
Mostly just documentation changes.

However, I do have one extra request: can you update the [`IDENTIFIER_KEYS` variable in `_dispatcher.py`](https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py#L53-L63) to include `group` and `day_obs`? It seems to have been overlooked when they were implemented.

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
timj: Sorry. I don't know what you mean. The default is `lsst.dm`.

kfindeisen: `__init__` assigns "lsst.dm" if nothing supersedes that. I meant "should something be assigned to `namespace` here, where it's declared?"

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
timj: Yes, I intended `None` there because that was the originally default for no fields and is what the dispatcher wants (it explicitly checks for None). I had not considered merging but that seems like a reasonable thing to do since these are not hierarchical.

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
timj: Ah. Okay. We could do that. Generally the class properties are there for docstrings and I'm not sure why Nate didn't set the default there instead, although a class property and an instance property can get confusing.

timj:
@kfindeisen please take another look. Completely rewritten. Thanks for the comments.

kfindeisen (approved):
Thanks, looks clear.

PR #229, opened 2024-03-24
DM-43477: Create pipeline yaml file to create calexp summary metrics from existing calexps' summary stats

jrmullaney:
Enables the production and writing of metrics generated from an existing calexp's summary statistics.

### DIFF
--- a/pipelines/calexpMetrics.yaml
+++ b/pipelines/calexpMetrics.yaml
@@ -1,0 +1,8 @@
+description: |
+  Metrics created from the analysis of calibrated exposures.
+tasks:
+  calexpSummary:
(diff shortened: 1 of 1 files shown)


jrmullaney:
Following the merge of DM-43231, it was decided that it would be good if one could make matrics from existing calexps, rather than having to create new calexps to make the metrics. I don't think this is possible without a pipeline or config file (i.e., it can't be done with a pipetask `-c` command line parameter), since the calexpMetrics.py analysis tool needs to be imported.

Note the namechange, though: I am no longer calling it `core` (or even `extended`) as it shouldn't be run automatically or semi-autoatically.

PR #233, opened 2024-04-08
DM-43681: Add calexpMetrics atool to AP metrics afterburner

ebellm:


pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -105,12 +105,20 @@ tasks:
       atools.interpolatedTemplatePsfSize.produce.plot.zAxisLabel: pixels
       python: |
         from lsst.analysis.tools.atools import DiffimSpatialMetricsHistPlot, DiffimSpatialMetricsInterpolatePlot
+  initialPviCore:
+    class: lsst.analysis.tools.tasks.CalexpSummaryAnalysisTask
+    config:
+      atools.calexpSummaryMetrics: CalexpSummaryMetrics
+      connections.data: initial_pvi.summaryStats
+      connections.outputName: initialPviSummary
```
isullivan, line 113: Add:
`connections.fakesType: parameters.fakesType`

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -105,12 +105,20 @@ tasks:
       atools.interpolatedTemplatePsfSize.produce.plot.zAxisLabel: pixels
       python: |
         from lsst.analysis.tools.atools import DiffimSpatialMetricsHistPlot, DiffimSpatialMetricsInterpolatePlot
+  initialPviCore:
+    class: lsst.analysis.tools.tasks.CalexpSummaryAnalysisTask
+    config:
+      atools.calexpSummaryMetrics: CalexpSummaryMetrics
```
isullivan, line 111: Since we call these PVIs now, consider renaming everything that references a `calexp` here. You probably should leave `CalexpSummaryAnalysisTask` as-is since I imagine that is used elsewhere, but both names in `atools.calexpSummaryMetrics: CalexpSummaryMetrics` should change.

isullivan (approved):
Looks good, but please rename the metrics to avoid the old `calexp` name, and add the `fakesType` to the connections.

pipelines/apDetectorVisitQualityCore.yaml
ebellm: We don't need to duplicate either the metric or the task, actually--retargeting the `outputName` below is sufficient that all of the downstream tools only see `pvi`, not `calexp`.  Since the underlying `summaryStats` input is the same in both cases, we avoid unnecessary code duplication.

isullivan: OK, but at least rename `atools.calexpSummaryMetrics` to `pviSummaryMetrics`. That name does make it through to the final metrics .json file, and is how it will appear in chronograf.

pipelines/apDetectorVisitQualityCore.yaml
ebellm: Unless I don't understand something about how analysis tools works (possible!), I don't think this is needed: Unlike the other Task inputs in this pipeline, the `initial_pvi` dataset type is upstream of all fakes data products.

isullivan: Ah, in that case you are right and it should not be needed.

pipelines/apDetectorVisitQualityCore.yaml
ebellm: Ah, I see.  I didn't realize the dataset type didn't reflect what was in the .json payload.  Confirmed the fix (I called it `initialPviSummaryMetrics`).

PR #236, opened 2024-04-12
DM-39256: Add AB1 and ABF1 metrics

cmsaunders:


python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -642,3 +644,108 @@ def setDefaults(self):
...
+        self.process.filterActions.brightStarsTot = DownselectVector(vectorKey="astromDiffTot")
+        self.process.filterActions.brightStarsTot.selector = RangeSelector(
+            vectorKey="mags", minimum=17, maximum=21.5
+        )
+
+        # Calculate median and sigmaMad
+        self.process.calculateActions.AB1_RA = RmsAction(vectorKey="brightStarsRA")
+        self.process.calculateActions.ABF1_RA = FracThreshold(
+            vectorKey="brightStarsRA", threshold=self.AB2, op="gt", relative_to_median=True, percent=True
+        )
```
jrmullaney, line 716: As agreed, this should go in a `finalize` block so that any runtime configs are picked-up (applies to the next few lines that use `self.AB2`, too).

Separately...looking at FracThreshold - doesn't it only return the fraction literally `gt` the threshold (so a large _negative_ separation will not be included in the fraction)? Since RA and Dec separation can have +ve and -ve values, wouldn't we want the fraction of abs(separation) > threshold?

python/lsst/analysis/tools/atools/refCatMatchPlots.py
```diff
@@ -642,3 +644,108 @@ def setDefaults(self):
...
+    """Calculate the AB1 and ABF1 metrics from the difference between the
+    target and reference catalog coordinates.
+    """
+
+    parameterizedBand = Field[bool](
+        doc="Does this AnalysisTool support band as a name parameter", default=False
+    )
+    AB2 = Field[float](
+        doc="Value by which no more than ABF1% of measurements shall deviate from the mean.",
+        default=20,
```
jrmullaney, line 659: I was a bit confused by this docstring since from reading the code, it looks as though `AB2` is used when calculating what percent bright stars have a ra/dec/tot astrometry difference greater than AB2 (with AB2 expressed in milliarcseconds). Have I interpreted the code correctly, or am I way off the mark? If the former, then I think this docstring needs re-writing.

jrmullaney (changes requested):
Please see in-line comments.

python/lsst/analysis/tools/atools/refCatMatchPlots.py
cmsaunders: Yes, definitely. I'm annoyed at myself for missing this, as I already dealt with this on another metric! I think I have to add an option to `FracThreshold` to take the absolute value.

python/lsst/analysis/tools/atools/refCatMatchPlots.py
cmsaunders: You have interpreted it correctly. The statement from the SRD is "No more than ABF1 % of the measurements will deviate by more than AB2 milliarcsec from the mean", so this was my attempt to invert the statement to describe AB2. 
How about "Threshold at which to calculate the percentage of measurements (the ABF1 metric) that are outliers from the mean."?

jrmullaney: How about: "Separation in milliarcseconds that is used when calculating the ABF1 metric. ABF1 is the fraction of sources whose astrometric separation-from-reference is greater than SB2 milliarcseconds from the mean."?

cmsaunders: I went with yours with some small changes.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -0,0 +1,115 @@
...
+    pass
+
+
+class SourceObjectTableAnalysisTask(AnalysisPipelineTask):
+    ConfigClass = SourceObjectTableAnalysisConfig
+    _DefaultName = "sourceTableVisitAnalysis"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+
```
jrmullaney, line 63: To get a more complete plotInfo box on the plot, suggest
```
dataId = butlerQC.quantum.dataId
plotInfo = self.parsePlotInfo(inputs, dataId)
```
is added here and the run command on line 114 is changed to:
`outputs = self.run(data=allCat, bands=band, plotInfo=plotInfo)`
.

jrmullaney (approved):
I've added a few suggestions for improvement, but I'm happy to sign off on this.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -0,0 +1,115 @@
...
+        doc="Table of associated sources",
+        name="{associatedSourcesInputName}",
+        storageClass="ArrowAstropy",
+        multiple=True,
+        deferLoad=True,
+        dimensions=("instrument", "skymap", "tract"),
+    )
+
+    refCat = ct.Input(
+        doc="Catalog of positions to .",
```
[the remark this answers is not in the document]
cmsaunders: Looking at this file while fixing this docstring, I realized the columns used for RA and Dec for the reference catalog need to be configurable, so I added that.

PR #237, opened 2024-04-12
DM-42927: Update cp_verify connections/classes/outputs for analysis_tools

czwa:
This updates the handling of calibration products.  I've renamed some classes and files to be more informative.  This also updates the `GridPlot` code to fix the column-vs-column plotting.


pipelines/cpCore.yaml
```diff
@@ -0,0 +1,280 @@
...
+      # atools.defectsColdPixels.produce.plot.addHistogram: true
+      # atools.defectsColdPixels.produce.plot.zAxisLabel: "Number of Cold Pixels"
+      # atools.defectsColdPixels.quantityKey: DEFECTS_DEFECTS_DIST_N_COLD
+      # atools.defectsColdPixels.unit: "---"
+      # atools.defectsBadColumns: CalibStatisticFocalPlanePlot
+      # atools.defectsBadColumns.produce.plot.addHistogram: true
+      # atools.defectsBadColumns.produce.plot.zAxisLabel: "Number of Hot Pixels"
+      # atools.defectsBadColumns.quantityKey: DEFECTS_DEFECTS_DIST_N_BAD_COLUMNS
+      # atools.defectsBadColumns.unit: "---"
+      python: |
```
sr525, line 242: Should all these commented out bits be here?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -86,12 +86,25 @@ class GridPlot(PlotAction):
...
         optional=True,
     )
+    xlabel = Field[str](
+        doc="String argument passed into fig.supxlabel() defining the figure x label.",
+        optional=True,
+    )
+    ylabel = Field[str](
+        doc="String argument passed into fig.supylabel() defining the figure y label.",
+        optional=True,
+    )
```
sr525, line 96: Can we make these xAxisLabel and yAxisLabel to bring them inline with other plots?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -109,19 +122,23 @@ def __call__(self, data: KeyedData, **kwargs) -> PlotResultType:
                         newData = {}
                         if val not in key:
                             continue
-                        namedKey = self.panels[index].plotElement.valsKey
-                        newData[namedKey] = data[key]
                         if xList is not None:
                             namedKey = self.panels[index].plotElement.xKey
                             newData[namedKey] = data[xList[i]]
+                            if key in xList:
```
sr525, line 128: I know that you didn't write most of this section but some comments in here would be really useful.

python/lsst/analysis/tools/atools/calibQuantityProfile.py
```diff
@@ -0,0 +1,358 @@
...
+        self.produce.plot.numCols = 2
+
+        # Values to group by to distinguish between data in differing panels
+        self.produce.plot.valsGroupBy = {
+            0: "PTC_COV_10",
+            1: "PTC_COV_20",
+            2: "PTC_COV_01",
+            3: "PTC_COV_02",
+            4: "PTC_COV_11",
+            5: None,
```
sr525, line 332: What does the 5: None do?

pipelines/cpCore.yaml
czwa: I've left commented blocks that have associated JIRA tickets.  I think in most of these cases, the commented code will work once those tickets merge (and hopefully un-comment these blocks).

python/lsst/analysis/tools/actions/plot/gridPlot.py
czwa: Documented as best I could since I'll likely need to work on this again.  There are multiple levels of indirection that make this tool difficult to use.

python/lsst/analysis/tools/atools/calibQuantityProfile.py
czwa: Grid plot needs to be able to generate an N*M grid of panels, but I only need 5, and 5 is unfortunately prime.  By setting this to `None`, no columns will match, and the sixth panel will be empty.

czwa (commented):
I hit the wrong button, so this is just me adding my additional comments.

PR #240, opened 2024-04-18
DM-43959: Add npatch field to BinnedCorr2Config

arunkannawadi:


rmjarvis:
I'll just mention that the term BinnedCorr2 isn't a thing in TreeCorr anymore (as of v5.0).  Probably should just use Corr2Config to avoid confusion going forward.

rmjarvis:
cf. https://rmjarvis.github.io/TreeCorr/_build/html/changes.html#api-changes

arunkannawadi:
We're still on `v4.3.3` as of this week (this might change next week if `rubin-env` gets an update). But now that I'm passing this around for almost all calls to `treecorr` methods, I could make this a `TreecorrConfig` or something like that.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -439,18 +448,30 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
         ra: Vector = data[self.colRa]  # type: ignore
         dec: Vector = data[self.colDec]  # type: ignore
 
-        treecorrKwargs = self.treecorr.toDict()
+        treecorr_config_dict = self.treecorr.toDict()
+
+        # Swap rng_seed with an rng instance in treecorr config.
+        rng = np.random.RandomState(treecorr_config_dict.pop("rng_seed"))
+        treecorr_config_dict["rng"] = rng
```
arunkannawadi, line 455: The `rng_seed` remains the same for all the runs for now, but a new `rng` is generated for each `tract, patch, band` combo. I suppose the actual patches themselves depends on the data (since it uses kmeans) and it's not the case where we get the same patch every single run, right?

rmjarvis (approved):
Code looks fine to me.  Should there be some new tests though?  Or is this just a functionality preserving change for now?

arunkannawadi:
We have tests for PSF residual calculations, but nothing for RhoStatistics now. In general, this has a lot less test coverage than many of the other DM repos, since it's mostly plot oriented. The plot I produced with the changes here (and activating them) is on the Jira ticket.

PR #241, opened 2024-04-26
DM-44002: Fix DatasetRef warning when using the analysis_tools reconstructor

jrmullaney:
Removed collections keyword from butler.get(ref,...) in tasks/reconstructor.py


jeffcarlin (approved):
Looks good to merge! I confirmed that it works (i.e., the warnings are gone).

PR #242, opened 2024-04-26
DM-44085: Generalize runQuantum to allow for arbitrary connections

arunkannawadi:


python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -60,6 +60,10 @@
 from ._interfaces import KeyedData, PlotTypes
 from ._metricMeasurementBundle import MetricMeasurementBundle
 
+# TODO: This rcParams modification is a temporary solution, hiding
+# a matplotlib warning indicating too many figures have been opened.
+# When DM-39114 is implemented, this should be removed.
+plt.rcParams.update({"figure.max_open_warning": 0})
 
 # TODO: This _plotCloser function assists in closing all open plots at the
```
leeskelvin, line 68: blank lines lint error below.

python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -435,19 +439,20 @@ def runQuantum(
...
-        outputs = self.run(data=data, plotInfo=plotInfo, skymap=skymap)
+        # We implicitly assume that 'data' has been defined, but do not have a
+        # corresponding input connection in the base class. Thus, we capture
+        # and re-raise the error with a more helpful message.
+        try:
+            # data has to be popped out to avoid duplication in the call to the
+            # `run` method.
+            inputData = inputs.pop("data")
+        except KeyError:
+            raise RuntimeError("'data' is a required input connection, but is not defined ")
```
leeskelvin, line 453: Should this raise end with a period, rather than whitespace?

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -42,55 +42,14 @@
 null_formatter = matplotlib.ticker.NullFormatter()
 
 
-def parsePlotInfo(dataId, runName, tableName, bands, plotName, SN):
```
leeskelvin, line 45: This is being moved from `plotUtils` into `diffimTaskDetectorVisitSpatiallySampledAnalysis`?

arunkannawadi:
I'll fix the linting errors while rebasing and reorganizing the commits.

python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -60,6 +60,10 @@
 from ._interfaces import KeyedData, PlotTypes
 from ._metricMeasurementBundle import MetricMeasurementBundle
 
+# TODO: This rcParams modification is a temporary solution, hiding
+# a matplotlib warning indicating too many figures have been opened.
+# When DM-39114 is implemented, this should be removed.
+plt.rcParams.update({"figure.max_open_warning": 0})
```
arunkannawadi, line 66: This gets run exactly once at the import time and even if the individual tasks override the `runQuantum` method.

python/lsst/analysis/tools/tasks/ccdVisitTableAnalysis.py
```diff
@@ -87,24 +87,3 @@ class CcdVisitTableAnalysisConfig(AnalysisBaseConfig, pipelineConnections=CcdVis
...
     _DefaultName = "ccdVisitTableAnalysisTask"
-
-    def runQuantum(self, butlerQC, inputRefs, outputRefs):
-        # Docs inherited from base class.
-        inputs = butlerQC.get(inputRefs)
-        dataId = butlerQC.quantum.dataId
-        plotInfo = self.parsePlotInfo(inputs, dataId)
-        data = self.loadData(inputs["data"])
-        skymap = None if "skymap" not in inputs.keys() else inputs["skymap"]
-        camera = None if "camera" not in inputs.keys() else inputs["camera"]
```
leeskelvin, line 98: So, if I read this correctly, we no longer need a special carve-out for `camera` here - is that correct? I see the base `runQuantum` has a carve-out for `skymap`, but not for `camera`.

python/lsst/analysis/tools/actions/plot/plotUtils.py
leeskelvin: Ah, I suppose this is being removed to defer to the main `parsePlotInfo` inside `analysis/tools/interfaces/_task.py` instead?

arunkannawadi: No, this is a function that's not called from anywhere. Instead, its namesake which a method of `AnalysisPipelineTask` is the one that's used everywhere. I'm removing it to avoid any confusion.

python/lsst/analysis/tools/tasks/ccdVisitTableAnalysis.py
arunkannawadi: In this ticket, even the base class shouldn't have a special carve-out for `skymap`. Defining connections should be sufficient, and no specializion of the `runQuantum` method is required. This is the main motivation for this ticket, because it was causing much confusion among the SciUnit members.

leeskelvin: Should we therefore also remove the special carve out for `skymap` here too?

https://github.com/lsst/analysis_tools/blob/1ca23f87940ab22fe96e2fce7408de272e41e4df/python/lsst/analysis/tools/interfaces/_task.py#L446

Or is this needed elsewhere?

python/lsst/analysis/tools/interfaces/_task.py
```diff
@@ -435,19 +440,20 @@ def runQuantum(
...
-        # When DM-39114 is implemented, this should be removed.
-        plt.rcParams.update({"figure.max_open_warning": 0})
         inputs = butlerQC.get(inputRefs)
         dataId = butlerQC.quantum.dataId
         plotInfo = self.parsePlotInfo(inputs, dataId)
-        data = self.loadData(inputs["data"])
-        if "skymap" in inputs.keys():
-            skymap = inputs["skymap"]
-        else:
-            skymap = None
```
arunkannawadi, line 449: `skymap` carve-out is being removed here.

python/lsst/analysis/tools/tasks/ccdVisitTableAnalysis.py
arunkannawadi: The PR removes this. I made an inline comment just now.

leeskelvin (approved):

leeskelvin:
This looks good, and captures a number of good QoL improvement! :+1:

PR #244, opened 2024-04-29
DM-43898: Create PlotInfo analysis tools plot element

jrmullaney:
Added a PlotInfoElement for use in element-wise analysis tools plot creation. Also made changes to GridPlot to accommodate the new PlotInfoElement.

### DIFF
--- a/python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
+++ b/python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
@@ -1,0 +1,78 @@
+# This file is part of analysis_tools.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ("PlotInfoElement",)
+
+from typing import TYPE_CHECKING, Any, Mapping
+
+import matplotlib.transforms as transforms
+
+from ....interfaces import PlotElement
+from ..plotUtils import parsePlotInfo
+
+if TYPE_CHECKING:
+    from matplotlib.axes import Axes
+
+
+class PlotInfoElement(PlotElement):
+    """Adds information to a plot, such as run collection name,
+    datasets used, dataset type, and tract/visit IDs.
+
+    Notes
+    -----
+    Converts a plotInfo dict to a string and creates a textbox to display it.
+
+    """
+
+    def __call__(self, ax: Axes, figureInfo: Mapping[str, Any], **kwargs) -> Axes:
+        """Draw a text box that displays useful information about the plot.
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`
+            Axes to plot on.
+        figInfo: `dict`
+            A dictionary containing information about the figure on which the
+            element is being placed.  It should have the following keys:
+
+            ``"figsize"``
+                The width and height of the figure in inches
+                (`tuple` of two `float`s)
+            ``"dpi"``
+                The dpi of the figure (`float`)
+
+        Notes
+        -----
+        plotInfoElement is added to the figure in-place, so no parameter is
+        returned.
+        """
+        plotInfo = kwargs["plotInfo"]
+        plotInfoText = parsePlotInfo(plotInfo)
+
+        # Ensures consistent spacing between the plotName and plotInfoText:
+        dpi_scale_trans = transforms.Affine2D().scale(figureInfo["dpi"])
+        trans = ax.transAxes + transforms.ScaledTranslation(0.0, -0.28, dpi_scale_trans)
+
+        ax.set_axis_off()
+        ax.text(0, 1.0, plotInfo["plotName"], fontsize=7, ha="left", va="bottom")
+        ax.text(0, 1.0, plotInfoText, fontsize=6, alpha=0.6, ha="left", transform=trans)
--- a/python/lsst/analysis/tools/actions/plot/gridPlot.py
+++ b/python/lsst/analysis/tools/actions/plot/gridPlot.py
@@ -26,6 +26,7 @@ __all__ = ("GridPlot", "GridPanelConfig")
 from typing import TYPE_CHECKING
 
 import matplotlib.pyplot as plt
+import numpy as np
 from lsst.pex.config import Config, ConfigDictField, DictField, Field, ListField
 from lsst.pex.config.configurableActions import ConfigurableActionField
 from matplotlib.gridspec import GridSpec
@@ -42,10 +43,11 @@ class GridPanelConfig(Config):
     )
     title = DictField[str, str](
         doc="String arguments passed into ax.set_title() defining the plot element title.",
+        optional=True,
     )
     titleY = Field[float](
         doc="Y position of plot element title.",
-        default=None,
+        optional=True,
     )
 
 
@@ -64,6 +66,14 @@ class GridPlot(PlotAction):
     numCols = Field[int](
         doc="Number of columns.",
         default=1,
+    )
+    width_ratios = ListField[float](
+        doc="Width ratios",
+        optional=True,
+    )
+    height_ratios = ListField[float](
+        doc="Height ratios",
+        optional=True,
     )
     xDataKeys = DictField[int, str](
         doc="Dependent data definitions. The key of this dict is the panel ID. The values are keys of data "
@@ -98,6 +108,18 @@ class GridPlot(PlotAction):
     def __call__(self, data: KeyedData, **kwargs) -> PlotResultType:
         """Plot data."""
         fig = plt.figure(figsize=self.figsize, dpi=self.dpi)
+        figureInfo = {"figsize": fig.get_size_inches(), "dpi": fig.get_dpi()}
+
+        if self.height_ratios is None:
+            height_ratios = np.ones(self.numRows) / self.numRows
+        else:
+            height_ratios = self.height_ratios / np.sum(self.height_ratios)
+
+        if self.width_ratios is None:
+            width_ratios = np.ones(self.numCols) / self.numCols
+        else:
+            width_ratios = self.width_ratios / np.sum(self.width_ratios)
+
         if self.suptitle is not None:
             fig.suptitle(**self.suptitle)
         if self.xAxisLabel is not None:
@@ -105,7 +127,14 @@ class GridPlot(PlotAction):
         if self.yAxisLabel is not None:
             fig.supylabel(self.yAxisLabel)
 
-        gs = GridSpec(self.numRows, self.numCols, figure=fig)
+        # TODO: See DM-44283:Add subplot_mosaic functionality to plotElements
+        gs = GridSpec(
+            self.numRows,
+            self.numCols,
+            figure=fig,
+            height_ratios=height_ratios,
+            width_ratios=width_ratios,
+        )
 
         # Iterate over all of the plots we'll make:
         for row in range(self.numRows):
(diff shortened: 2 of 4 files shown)


python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -90,16 +103,48 @@ class GridPlot(PlotAction):
...
+            self.numRows + addPlotInfoRow,
+            self.numCols,
+            figure=fig,
+            height_ratios=height_ratios,
+            width_ratios=width_ratios,
+        )
+
+        if not includesPlotInfoPanel:
+            ax = fig.add_subplot(gs[0, :])
+            _ = PlotInfoElement().makeElement(ax=ax, fig=fig, **kwargs)
```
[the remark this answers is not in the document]
jrmullaney: Ok. It's used to calculate a scale-independent spacing between the plotInfo title and the remaining text. I can see if there's another way to do it. What's the justification for not wanting specific elements (as opposed to all) being passed a fig object?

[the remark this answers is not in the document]
natelust: I think it is a violation of separation of concerns and encapsulation. I think it will also encourage others to look at this, also pass the figure object, and have the elements make changes to the figure. If all you need in the dpi spacing or something like that either pass that, or possibly look to see if that is context we think other elements may need to know.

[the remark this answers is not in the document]
jrmullaney: Thanks - makes total sense. I'll pass the fig scale transform instead, which is all that it's needed for.

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Returns
+        -------
+        fig : `matplotlib.figure.Figure`
+            The figure with the information added.
+        """
+        # TO DO: This function is a near copy-paste of addPlotInfo
+        #        from plotUtils, but which only builds the plot info
+        #        text, rather than adding it to a figure. Suggest the
+        #        text-build is separated-out from addPlotInfo in plotUtils.
```
leeskelvin, line 60: This is an important TODO, and something we might want to consider now. Adding logic to parse a `plotInfo` dict here duplicates logic already inside `addPlotInfo` in `plotUtils`, and introduces the possibility of a divergence down the line.

Instead of duplicating code, it seems to me that the better approach would be (as you rightly point out in this TODO) to separate out the dict-parsing logic inside `addPlotInfo` and drop it into its own (recently freed-up name) `parsePlotInfo`. That should allow it to be imported both here an inside the remainder of `addPlotInfo` as appropriate. I don't think such an exercise is beyond the scope of this ticket - what do you think?

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
```diff
@@ -0,0 +1,146 @@
...
+        selectionPrefix = "Selection: "
+        for key, value in plotInfo.items():
+            if "SN" in key or "S/N" in key:
+                snKeys.append(key)
+            elif "Mag" in key:
+                magKeys.append(key)
+            elif key.startswith(selectionPrefix):
+                selectionKeys.append(key)
+        # Add S/N and mag values to label, if present.
+        # TODO: Do something if there are multiple sn/mag keys. Log? Warn?
```
leeskelvin, line 94: I realize this is copied over from the existing code, so I'm not expecting you to address it here necessarily. However, what I would do were it me is to add an optional `logger` argument into this function which is a logging instance. Then, inside this function where appropriate, add `if logger` blocks which then call `logger.info` or `logger.warning` etc when one of these triggers is `True`.

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
```diff
@@ -0,0 +1,146 @@
...
+            ``"dpi"``
+                The dpi of the figure (`float`)
+
+        Returns
+        -------
+        ax : `~matplotlib.axes.Axes`
+            Updated axes.
+        """
+        plotInfo = kwargs["plotInfo"]
+        plotInfoText = self._buildPlotInfoText(plotInfo)
```
leeskelvin, line 133: Can be replaced with a call to the imported function once/if the above code is moved back into a new `parsePlotInfo` function inside `plotUtils`.

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
```diff
@@ -0,0 +1,146 @@
...
+
+        ax.set_axis_off()
+        ax.text(0, 1.0, plotInfo["plotName"], fontsize=7, ha="left", va="bottom")
+        ax.text(0, 1.0, plotInfoText, fontsize=6, alpha=0.6, ha="left", transform=trans)
+
+        # QUERY: Other plotElements return "data", but "data"
+        #        isn't passed to this plotElement "ax" here is a
+        #        placeholder, since it needs to return something.
+        #        Should we use "None"?
+        return ax
```
leeskelvin, line 146: We don't necessarily need to pass data/anything out, do we? If we do need to pass something out, perhaps the parsed plot info would be a good data point to have?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -65,14 +68,24 @@ class GridPlot(PlotAction):
         doc="Number of columns.",
         default=1,
     )
+    width_ratios = ListField[float](
+        doc="width ratios",
+        optional=True,
+    )
+    height_ratios = ListField[float](
+        doc="height ratios",
```
leeskelvin, line 76: "Width ratios."
"Height ratios."

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -65,14 +68,24 @@ class GridPlot(PlotAction):
...
+        "The values are keys of data to plot (comma-separated for multiple) where "
+        "each key may be a subset of a full key.",
         default={},
     )
     valsGroupBy = DictField[int, str](
-        doc="Independent data definitions. The key of this dict is the panel ID. The values are keys of data "
-        "to plot (comma-separated for multiple) where each key may be a subset of a full key.",
+        doc="Independent data definitions. The key of this dict is the panel ID. "
+        "The values are keys of data to plot (comma-separated for multiple) where "
+        "each key may be a subset of a full key.",
```
leeskelvin, line 88: Other than moving the sentence to a new line, did something else change here?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -90,16 +103,49 @@ class GridPlot(PlotAction):
...
+
+        if self.height_ratios is None:
+            height_ratios = np.ones(self.numRows) / self.numRows
+        else:
+            height_ratios = self.height_ratios / np.sum(self.height_ratios)
+
+        if self.width_ratios is None:
+            width_ratios = np.ones(self.numCols) / self.numCols
+        else:
+            width_ratios = self.width_ratios / np.sum(self.width_ratios)
```
leeskelvin, line 116: We probably need a couple of new validations down in `def validate` that verify that the length of supplied `height_ratios`/`width_ratios` (if given) matches the `numRows`/`numCols` config values.

leeskelvin: What happens now if you try to construct your test plot with a plot Info, but pass too many `height_ratios`? Does it fail it a meaningful error?

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
leeskelvin: But if nothing, this could just be a method which edits in-place.

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -90,16 +103,49 @@ class GridPlot(PlotAction):
...
+        includesPlotInfoPanel = False
+        addPlotInfoRow = 1
+        for panel in self.panels.values():
+            if isinstance(panel.plotElement, PlotInfoElement):
+                includesPlotInfoPanel = True
+                addPlotInfoRow = 0
+                break
+
+        if not includesPlotInfoPanel:
+            height_ratios = np.append([1.0], 20.0 * height_ratios)
```
leeskelvin, line 127: What's the purpose of this? Is it to reserve some small fraction of space at the top of a plot if a plot info is _not_ given?

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -109,19 +155,22 @@ def __call__(self, data: KeyedData, **kwargs) -> PlotResultType:
...
+                        _ = self.panels[index].plotElement(
+                            data=newData, ax=ax, figureInfo=figureInfo, **kwargs
+                        )
 
                 if self.panels[index].title is not None:
                     ax.set_title(**self.panels[index].title, y=self.panels[index].titleY)
 
         plt.tight_layout()
-        fig.show()
+        # fig.show() # Should this be here?
```
leeskelvin, line 173: No, I think not! Well spotted...!

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -90,16 +103,49 @@ class GridPlot(PlotAction):
...
+            self.numRows + addPlotInfoRow,
+            self.numCols,
+            figure=fig,
+            height_ratios=height_ratios,
+            width_ratios=width_ratios,
+        )
+
+        if not includesPlotInfoPanel:
+            ax = fig.add_subplot(gs[0, :])
+            _ = PlotInfoElement().makeElement(ax=ax, figureInfo=figureInfo, **kwargs)
```
leeskelvin, line 141: Something about this doesn't feel quite right, i.e., a special carve-out for a `PlotInfoElement` (here and below). Perhaps the `data` for a `PlotInfoElement` is the `figureInfo`, which would then bring it into line with how other elements have been designed - i.e., "data input needed to make the element work".

leeskelvin: This may require us to re-think how we generate gridspecs somewhat though, i.e., we probably want the user to be able to specify the entire top-row in a gridplot should be reserved for use by a plot info element, rather than hard-coding it here.

python/lsst/analysis/tools/actions/plot/gridPlot.py
leeskelvin: If so, please add a brief in-line comment above with your rationale. However, pending a discussion below of a potential re-write of how we handle plot info elements, this may end up disappearing in the rewrite.

leeskelvin (commented):
Some initial thoughts which I hope are helpful. I agree that moving some of the plot info dict parsing login into a new `parsePlotInfo` function inside `plotUtils` probably makes a ton of sense. I see now on the Jira ticket that that was your recommendation also, so I wholeheartedly support that.

Some further comments here on whether or not we want to be handling plot info elements in a special way, which goes against the ethos of the "plot elements" system somewhat. It may be a necessary evil here however, considering that plot info is a somewhat special case.

python/lsst/analysis/tools/actions/plot/gridPlot.py
jrmullaney: No, I believe this was just a black edit.

leeskelvin: Hm, interesting, black doesn't change this for me. At least, not in VSCode with this in my `settings.json`:
```
    "black-formatter.args": [
        "--line-length",
        "110",
    ],
```

python/lsst/analysis/tools/actions/plot/elements/plotInfoElement.py
```diff
@@ -0,0 +1,146 @@
...
+        """Add useful information to the plot.
+
+        Parameters
+        ----------
+            A dictionary of the plot information.
+
+        Returns
+        -------
+        fig : `matplotlib.figure.Figure`
+            The figure with the information added.
```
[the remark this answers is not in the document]
jrmullaney: Now moved to `plotUtils.py` with updated docstrings.

python/lsst/analysis/tools/actions/plot/gridPlot.py
jrmullaney: This is fixed now - keeping the longer lines. Don't know what happened there!

python/lsst/analysis/tools/actions/plot/gridPlot.py
jrmullaney: Validation added.

python/lsst/analysis/tools/actions/plot/gridPlot.py
jrmullaney: For now, this has been removed in lieue of developing a non-carve-out method of adding plotInfo if/when not explicitly specified.

python/lsst/analysis/tools/actions/plot/gridPlot.py
jrmullaney: This "auto-add plotInfo" feature has been removed (see "in lieue" above).

python/lsst/analysis/tools/actions/plot/plotUtils.py
```diff
@@ -285,6 +283,26 @@ def addPlotInfo(fig: Figure, plotInfo: Mapping[str, str]) -> Figure:
...
+    plotInfo : `dict`
+        A dictionary of the plot information.
+
+    Returns
+    -------
+    fig : `matplotlib.figure.Figure`
+        The figure with the information added.
+    """
+    fig.text(0.01, 0.99, plotInfo["plotName"], fontsize=7, transform=fig.transFigure, ha="left", va="top")
+    infoText = parsePlotInfo(plotInfo)
```
leeskelvin, line 305: Much cleaner!

leeskelvin:
Thanks @jrmullaney, I think this is looking great. This ticket now focuses solely on adding the new `PlotInfoElement`, and modifying `GridPlot` to be able to work with non-equal-area grid elements, such as a `PlotInfoElement`.

As we discussed, more work probably needs to be done to move `GridSpec` onto use of `subplot_mosaic`, to allow the user complete control over placement of plot elements. Did you set up a follow-up ticket for that? If so, please do add a `# TODO: See DM.....` in-line comment on this ticket to indicate that future effort will take place inside `GridSpec`.

Were it up to me, I'd approve this PR at this time. However, I see you've asked for @natelust and @sr525 for a review also, so I'll leave this open for now for them to leave their own comments if they wish. Thanks again for putting this together; it'll be super useful!

jrmullaney:
Thanks, @leeskelvin . I've created that new ticket (DM-44283) and added a TODO referring to it. I've also removed `atools/gridPlotTest.py` from the PR as it shouldn't be included in a merge. I'll run Jenkins now and wait for @natelust and @sr525 to sign-off or comment.
Thanks for your help on this!

leeskelvin (approved):

PR #245, opened 2024-04-30
DM-44085-hotfix: Make skymap optional in makePlot for ScatterPlotWithTwoHists

arunkannawadi:


arunkannawadi:
`generateSummaryStats` requires `skymap` argument and cannot handle it being `None`. However, for the visit-level plots, `addSummaryPlot` had been set to `False`, hence not calling it in the first place. But unlike `focalPlanePlot` which requires `camera`, this plotting is generic and should not require any argument that's not absolutely necessary.

leeskelvin:
PS - I would recommend a safety Jenkins with `ci_hsc` added into the build products prior to merge.

PR #246, opened 2024-05-07
DM-40203: Add a method for creating all sky plots of metrics in analysis_tools

jtmccann:


python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -0,0 +1,309 @@
...
+class WholeSkyPlot(PlotAction):
+    """Plots the on sky distribution of a parameter.
+
+    Plots the values of the parameter given for the z axis
+    according to the positions given for x and y. Optimised
+    for use with RA and Dec. Also calculates some basic
+    statistics and includes those on the plot.
+    """
+
+    xAxisLabel = Field[str](doc="Label to use for the x axis.")
```
sr525, line 53: These could have defaults

python/lsst/analysis/tools/scripts/allSkyProj.py
jtmccann, line 1: This is the draft of the plotting script and should not be used. There's some elements in here that I'd like to use on future tickets, so I didn't think it was appropriate to delete this file just yet.

python/lsst/analysis/tools/atools/wholeSkyPlotTool.py
```diff
@@ -0,0 +1,90 @@
...
+            "corners",
+            "tract",
+        ]
+        self.produce.plot = WholeSkyPlot()
+
+    def finalize(self):
+        # Loop over bands to explicitly define names of band-specific metrics.
+        for key in self.keysWithBand:
+            self.plotKeys.extend([key.format(band=band) for band in self.bands])
+        # This should be in validate, but for some reason finalize gets called
```
jtmccann, line 73: This section here with the ValueErrors is the only thing I might want to change... I'm not sure why validate is running after finalize.

python/lsst/analysis/tools/atools/wholeSkyPlotTool.py
```diff
@@ -0,0 +1,90 @@
...
+        # after.
+        if len(set(self.plotKeys)) != len(self.plotKeys):
+            raise ValueError(f"{self.plotKeys=} has duplicate keys.")
```
jtmccann, line 76: This section here with the ValueErrors is the only thing I might want to change... I'm not sure why validate is running after finalize.

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -0,0 +1,318 @@
...
+        -------
+        text : `str`
+            A string containing the 5 tracts with the largest outlier values.
+        """
+        text = "Outlier values: "
+        if len(outlierInds) > 0:
+            outlierValues = np.array(values)[outlierInds]
+            outlierTracts = np.array(tracts)[outlierInds]
+            # Sort values in descending (-) order discounting NaNs.
+            maxInds = np.argsort(-outlierValues)
```
sr525, line 110: Would it be better to do this by absolute value of the outlier?

python/lsst/analysis/tools/atools/wholeSkyPlotTool.py
```diff
@@ -0,0 +1,96 @@
...
+        super().setDefaults()
+        self.prep.keysToLoad = [
+            "corners",
+            "tract",
+        ]
+        self.produce.plot = WholeSkyPlot()
+
+    def finalize(self):
+        # Prevent finalize from running twice.
+        if not self.produce.plot.plotKeys:
```
jtmccann, line 76: finalize is getting called twice when I run this on ci_hsc, so I added this if statement to make sure it only runs once. Not sure why this is happening on ci_hsc but not anything else.

PR #249, opened 2024-05-13
DM-42595: Make a whole sky table of metrics

sr525:
Tables that include data for the metrics across all of the tracts in the sky are useful for a variety of reasons, including plotting whole sky plots.

### DIFF
--- a/python/lsst/analysis/tools/tasks/makeMetricTable.py
+++ b/python/lsst/analysis/tools/tasks/makeMetricTable.py
@@ -1,0 +1,206 @@
+# This file is part of analysis_tools.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = (
+    "MakeMetricTableConfig",
+    "MakeMetricTableTask",
+)
+
+
+import lsst.pipe.base as pipeBase
+import numpy as np
+from astropy.table import Table
+from lsst.pipe.base import connectionTypes as ct
+from lsst.skymap import BaseSkyMap
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class MakeMetricTableConnections(
+    AnalysisBaseConnections,
+    dimensions=("skymap",),
+    defaultTemplates={"metricBundleName": "objectTableCore_metrics"},
+):
+    data = ct.Input(
+        doc="Metric bundle to read from the butler",
+        name="{metricBundleName}",
+        storageClass="MetricMeasurementBundle",
+        deferLoad=True,
+        dimensions=(
+            "skymap",
+            "tract",
+        ),
+        multiple=True,
+    )
+
+    skymap = ct.Input(
+        doc="The skymap that covers the tract that the data is from.",
+        name=BaseSkyMap.SKYMAP_DATASET_TYPE_NAME,
+        storageClass="SkyMap",
+        dimensions=("skymap",),
+    )
+
+    metricTable = ct.Output(
+        doc="A summary table of all metrics by tract.",
+        name="{metricBundleName}Table",
+        storageClass="ArrowAstropy",
+        dimensions=("skymap",),
+    )
+
+
+class MakeMetricTableConfig(AnalysisBaseConfig, pipelineConnections=MakeMetricTableConnections):
+    pass
+
+
+class MakeMetricTableTask(AnalysisPipelineTask):
+    """Turn metric bundles which are per tract into a
+    summary metric table.
+
+    TO DO: DM-44485 make sure this works for visit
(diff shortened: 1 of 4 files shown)


python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -0,0 +1,203 @@
...
+        # Make a list of tracts
+        tracts = []
+        for data in inputRefs.data:
+            tracts.append(data.dataId["tract"])
+
+        metricBundles = []
+        for inputHandle in inputs["data"]:
+            metricBundles.append(inputHandle.get())
+
+        outputs = self.run(tracts, metricBundles, skymap)
```
arunkannawadi, line 106: Flyby review here. Could you move this data handling logic by overloading `_loadData` method and keep `runQuantum` the same from the base class? And keeping the signature of the `run` method the same as in the other tasks (i.e., `data`, `plotInfo` as arguments) would keep this in line with everything else. You might need to overload `parsePlotInfo` to return `None` to achieve that.

jtmccann:
Looks good to me!

python/lsst/analysis/tools/tasks/makeMetricTable.py
sr525: I am happy with the way it is, either way stuff will be different from elsewhere and I tried various options and this seemed to work the best.

python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -0,0 +1,203 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = (
+    "MakeMetricTableConfig",
+    "MakeMetricTableTask",
```
mrawls, line 25: Suggest explicitly adding tract to these names, e.g., `MakeMetricTableTractTask`

python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -0,0 +1,203 @@
...
+                    if fullName in metricsDict.keys():
+                        metricsDict[fullName].append(metricValue)
+                    else:
+                        values = [np.nan] * (len(metricsDict["tract"]) - 1)
+                        values.append(metricValue)
+                        metricsDict[fullName] = values
+                    metricNames.append(fullName)
+
+            # If a metric that existed for a previous tract does
+            # not exist for this one then add a nan
```
mrawls, line 165: I'd rather just skip it vs quietly making tables full of surprise nans, or at least log a warning

python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -0,0 +1,203 @@
...
+            for metricName in metricsDict.keys():
+                if metricName not in metricNames:
+                    metricsDict[metricName].append(np.nan)
+
+        tMetrics = pipeBase.Struct(metricTable=Table(metricsDict))
+        return tMetrics
+
+    def getTractCorners(self, skymap, tract):
```
mrawls, line 173: Can't this be simpler, as in plotUtils, e.g., `tractCorners = Box2D(tractInfo.getBBox()).getCorners()` ?

python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -0,0 +1,203 @@
...
+        for i, metricBundle in enumerate(metricBundles[1:]):
+            tract = tracts[i + 1]
+            metricsDict["tract"].append(tract)
+            corners = self.getTractCorners(skymap, tract)
+            metricsDict["corners"].append(corners)
+            metricNames = ["tract", "corners"]
+            for name, metrics in metricBundle.items():
+                for metric in metrics:
+                    fullName = f"{name}_{metric.metric_name}"
+                    metricValue = metric.quantity.value
```
mrawls, line 154: Please also keep track of metric units, either by writing Quantities to each table value or by somehow indicating in the column names what the appropriate unit

python/lsst/analysis/tools/tasks/makeMetricTable.py
mrawls: Or maybe `WholeSkyMetricTableTractTask` ? You call it `wholeSkyMetricTableOTC` in the pipeline

arunkannawadi: Yeah, what _is_ OTC?

sr525: I haven't added tract into the names yet because without doing the visit bit I am not sure shich bits are going to be tract level specific, I am hoping to also include which tract each visit overlaps, not sure how I am dealing with multiples yet but I think almost all of this code is applicable to both, except I still need to understand the AP data structures.

python/lsst/analysis/tools/tasks/makeMetricTable.py
sr525: I don't want any of the rows/columns to be different lengths. There needs to be something in the table cell if there is no entry.

python/lsst/analysis/tools/tasks/makeMetricTable.py
sr525: I don't think so? That doesn't do the wrapping that the plots need or the unit conversions.

python/lsst/analysis/tools/tasks/makeMetricTable.py
jtmccann: I see the name change in the pipeline file, and in light of your intentions to add visit level support I think this can be resolved. If you want to be even more explicit you might add something like "TODO: Add visit level support on DM-XXXXX" in the docstring.

jtmccann: I see the name change in the pipeline file, and in light of your intentions to add visit level support I think this can be resolved. If you want to be even more explicit you might add something like "TODO: Add visit level support on DM-XXXXX" in the docstring.

jtmccann (approved):
There's one optional comment of mine you can respond to above, but other than that everything looks good to me.

PR #251, opened 2024-05-14
DM-44302: Implement photometric repeatability metrics and plots in analysis_tools with calib_fluxes

jeffcarlin:
This ticket creates new metrics and plots that calculate photometric repeatability using `calibFlux` instead of `psfFlux`. The comparison between these is informative to the Photometric Calibration Science Unit in assessing different contributions to the error budget.

Implementing the focal plane plot required updating the `StellarPhotometricResidualsFocalPlane` class so that it will correctly apply config overrides.

Finally, I added formatting to the `histPlot` routine to make the reference_label formatting not have too many digits after the decimal.

### DIFF
--- a/pipelines/matchedVisitQualityCore.yaml
+++ b/pipelines/matchedVisitQualityCore.yaml
@@ -7,6 +7,10 @@ tasks:
       connections.outputName: matchedVisitCore
       atools.stellarPhotometricRepeatability: StellarPhotometricRepeatability
       atools.stellarPhotometricResiduals: StellarPhotometricResidualsFocalPlane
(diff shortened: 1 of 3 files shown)


erykoff (approved):
This is fine, but there is one part that makes me nervous, and that's the different s/n selection.  The psf flux will typically have much greater s/n than the calib flux (large aperture).  So the selection of stars is different for an equivalent s/n threshold.  I think that's why the sample plot in jira has fewer objects for calib flux.  Is it possible to make this separately configurable, and default to `psfFlux` in all cases for better apples-to-apples comparisons?

PR #253, opened 2024-05-16
DM-44349: Create a metric to count all vs "good" diaSources

jrmullaney:


python/lsst/analysis/tools/atools/diaSourceMetrics.py
```diff
@@ -30,26 +30,49 @@
...
+            actionA=self.process.calculateActions.numGoodDiaSources,
+            actionB=self.process.calculateActions.numAllDiaSources,
+        )
 
         # the units for the quantity (count, an astropy quantity)
-        self.produce.metric.units = {"numDiaSources": "ct"}
+        self.produce.metric.units = {
+            "numAllDiaSources": "ct",
+            "numGoodDiaSources": "ct",
+            "ratioGoodToAllDiaSources": "",
```
mrawls, line 70: Is an empty string allowed for a unit?

python/lsst/analysis/tools/atools/diaSourceMetrics.py
```diff
@@ -30,26 +30,49 @@
...
+class NumGoodDiaSourcesMetrics(AnalysisTool):
     """Calculate the number of DIA Sources that do not have known
-    bad/quality flags set to true.
+    bad/quality flags set to true, and also calculate the ratio of
+    counts of non-flagged sources to all sources.
     """
 
+    parameterizedBand: bool = False
+
+    applyDrpContext = Field[bool](doc="Apply the DRP context if analysing DRP DIA tables.", default=True)
```
mrawls, line 46: I think this and https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/actions/vector/selectors.py#L415 ... and all mentions of "DrpContext" throughout analysis_tools ...! can all go away now, because all the flag columns are simply named `pixelFlags_thing` and not `base_pixelFlags_flag_thing`.

python/lsst/analysis/tools/atools/diaSourceMetrics.py
```diff
@@ -30,26 +30,49 @@
...
-        self.produce.metric.units = {"numDiaSources": "ct"}
+        self.produce.metric.units = {
+            "numAllDiaSources": "ct",
+            "numGoodDiaSources": "ct",
+            "ratioGoodToAllDiaSources": "",
+        }
+
+    def finalize(self):
+        if self.applyDrpContext:
+            self.applyContext(DrpContext)
```
mrawls, line 75: Remove, assuming everything runs without it if you change the column name defaults to be `pixelFlags_thing` format.

tests/test_diaSourceMetrics.py
```diff
@@ -0,0 +1,50 @@
...
+from lsst.analysis.tools.contexts import DrpContext
+
+
+class DiaSourceTest(lsst.utils.tests.TestCase):
+    """Test to see if DiaSources are counted as expected
+    and a plot is generated without falling over."""
+
+    def setUp(self):
+        testFile = "tests/diaSourceTable_tract_test.ecsv"
+        self.data = Table.read(testFile)
```
mrawls, line 35: Please make a new test file that's called `assocDiaSrc_test.ecsv` or similar. I made this original one by loading up an instance of the data product in question with the Butler (in this case it's a `goodSeeingDiff_assocDiaSrc`) and writing out 100 lines or so to an ecsv file via astropy Table export. I just want to make sure we aren't accidentally testing against a different table schema, since the one you're using here is a DRP data product I use for another test.

python/lsst/analysis/tools/atools/diaSourceMetrics.py
jrmullaney: Once @erinleighh has pushed their changes to `GoodDiaSourceSelector`, I'll rebase and remove mention of DrpContext.

python/lsst/analysis/tools/atools/diaSourceMetrics.py
jrmullaney: I _think_ so, based on other atools, such as [this](https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/atools/deblenderMetric.py) (although I realise that's somewhat cargo-cult). I don't think there is any way round it, though, as non-empty strings have to be an astropy unit, but in some instances (such as angular area) an appropriate astropy unit doesn't exist.

python/lsst/analysis/tools/atools/diaSourceMetrics.py
jrmullaney: See reply above.

tests/test_diaSourceMetrics.py
jrmullaney: Done. I've left a comment on the ticket explaining what dataset I used, and from which repo and collection (for future reference and reproducability purposes).

mrawls (approved):
Looks great! Please coordinate with Erin as discussed, make sure a Jenkins run with ci_ap included is happy, and feel free to merge.

PR #256, opened 2024-05-21
DM-44342: Update flags in GoodDiaSourceSelector, fix DiaSourceTableTractTest, remove DRP Context.

erinleighh:


jrmullaney (approved):
This all looks fine to me.

Let's do a quick search for `DrpContext` once this is merged to make sure we've caught all instances; it seems that GitHub will only search the main branch, which is annoying! I've checked, and the changes I've made on DM-44349 work fine now without DrpContext, which is exactly what we wanted!

Thanks for getting on to this quickly; let me know when you've merged-to-main, then I'll rebase and re-push.

PR #257, opened 2024-05-22
DM-44484: Ensure all isolated catalogs are input to SourceObjectTableAnalysisTask and check for no sources.

erykoff:


cmsaunders (approved):
Looks good to me. I just noticed that most of the `defaultTemplates` are not actually used in the connections, and the one that is used `associatedSourcesInputName` is pretty unnecessary, since there are currently no other options that would work for it. You could get rid of these on this ticket, since you are already changing the connections, but that's definitely optional.

erykoff:
Since I've already run it through the full jenkins wringer I don't want to risk more connection changes.  😄

PR #259, opened 2024-05-29
DM-44565: Add coadd selectors to sourceTableVisitAnalysis

cmsaunders:


python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -85,11 +85,21 @@ class SourceObjectTableAnalysisConfig(
         dtype=str,
         default="r_dec",
     )
+    refCat_bands = pexConfig.ListField(
+        doc="Bands in refCat.",
+        dtype=str,
+        default=["u", "g", "r", "i", "z", "y"],
+    )
+    refCat_selectors = pexConfig.ListField(
+        doc="Remove objects for which these flags are true.",
```
jrmullaney, line 94: I think it needs to be made clear in the docstrings that  these two configs work together, i.e., that the refCat_selectors are combined with the refCat_bands above.

jrmullaney (approved):

PR #263, opened 2024-06-04
DM-43078: Integrate HealSparse property map metadata into analysis_tools visualizations

enourbakhsh:


python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -273,10 +274,23 @@ def makePlot(
...
                 # Make a concise human-readable label for the plot.
+                plotInfo["unit"] = "N/A"  # Unless overridden.
+                if hasattr(mapData, "metadata") and all(
+                    key in mapData.metadata for key in ["DESCRIPTION", "OPERATION", "UNIT"]
+                ):
+                    metadata = mapData.metadata
+                    plotInfo["description"] = metadata["DESCRIPTION"]
+                    plotInfo["operation"] = metadata["OPERATION"]
+                    if metadata["UNIT"]:
+                        plotInfo["unit"] = metadata["UNIT"]
```
erykoff, line 285: I think the logic here is not quite right.  If `metadata["UNIT"]` is set to `""` (empty string) then the unit really is `dimensionless` and not `N/A`.  So I think that (a) here should set `plotInfo["unit"] = "dimensionless"` if there's no unit; (b) below it should check for either `N/A` or `dimensionless` and not put the unit on the plot.  And that will create the right distinction in the plot title.

erykoff (approved):
Looks good, but a couple of small comments.  First, I'd like some feedback from the broader group on whether the units should be denoted with square brackets `[]` or round brackets `()`.  I prefer round brackets for units, but if there's consensus for square I won't fight.  Second, see my comments on `N/A` vs `dimensionless`.

enourbakhsh:
> Looks good, but a couple of small comments. First, I'd like some feedback from the broader group on whether the units should be denoted with square brackets `[]` or round brackets `()`. I prefer round brackets for units, but if there's consensus for square I won't fight. Second, see my comments on `N/A` vs `dimensionless`.

Thank you for the review @erykoff. The main reason I used square brackets `[]` instead of round brackets `()` is because the unit for `PsfMaglimPropertyMap` is `mag(AB)`. Using round brackets would have resulted in double parentheses on the plot, appearing as: `PSF maglim (mag(AB))`, which I find visually unappealing. Do you still prefer using round brackets? I can also set up a pool on #dm-science-pipelines and get feedback from the broader group as you suggested.

erykoff:
Oh that is very unfortunate, I don't know what astropy was thinking with that `mag(AB)`.  Sigh.  I would still open it up to the broader group.  Thanks!

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
enourbakhsh: Done!

PR #265, opened 2024-06-12
[WIPstart work on supporting multiple quantities in CalibAmpScatterTool

weatherhead99:
WIP


weatherhead99:
this now works and I've verified that the plots produced are identical to the original. Should we wish to change some plots to have multiple series in the calibAmpScatterTool it can now be done with the new ListField for quantityKey

czwa:
Can I ask that this be renamed according to the style guide (https://developer.lsst.io/work/flow.html#make-a-pull-request)?

I'm also somewhat hopeful that this ticket will magically fix another problem I'm having with these plot types, so I'm sorry it's taken so long to get to the review.

python/lsst/analysis/tools/actions/plot/gridPlot.py
```diff
@@ -84,10 +84,7 @@ class GridPlot(PlotAction):
         doc="Independent data definitions. The key of this dict is the panel ID. The values are keys of data "
         "to plot (comma-separated for multiple) where each key may be a subset of a full key.",
     )
-    figsize = ListField[float](
-        doc="Figure size.",
-        default=[8, 8],
-    )
+    figsize = ListField[float](doc="Figure size.", default=[8, 8], length=2)
```
czwa, line 87: Can this be expanded out to match the other fields formatting?

czwa (approved):
I'll let @natelust say if he has any concerns here, but this seems like an elegant fix to me.

python/lsst/analysis/tools/actions/plot/gridPlot.py
weatherhead99: @czwa I'm not sure I understand, do you just mean the code formatting? Or to change the other fields to further constrain their formatting?

czwa: Just code formatting, and I prefer the expanded "one element per line" version.

czwa:
I think this has been open long enough, so ping me after you've rebased, I'll do one final quick scan, and we can move ahead with getting this merged (without waiting for @natelust ).

PR #266, opened 2024-06-12
DM-44793: Get amp-to-amp offset pedestals from ISR metadata into Chronograf

enourbakhsh:


enourbakhsh:
> General comment/question: `visit` is used everywhere here, from the task name to the dimensions, but isr is at the `exposure` level. Is `visit` really correct here?

Thanks, @mfisherlevine. They will all be `exposure` once I push my new changes. Looking for the appropriate pipeline to put the task in.

pipelines/exposureQualityCore.yaml
```diff
@@ -0,0 +1,13 @@
+description: |
+  Tier 1 plots and metrics to assess exposure quality
+tasks:
+  analyzeAmpOffsetTaskCore:
+    class: lsst.analysis.tools.tasks.IsrAmpOffsetMetricsAnalysisTask
```
natelust, line 5: Should the task be named metrics since tools within it might contain plots?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+
+class IsrDetectorExposureMetricsAnalysisConfig(
+    AnalysisBaseConfig, pipelineConnections=IsrDetectorExposureMetricsAnalysisConnections
+):
+    pass
+
+
+class IsrAmpOffsetMetricsAnalysisTask(AnalysisPipelineTask):
```
natelust, line 49: Some comment above, should metrics be in the name?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
```diff
@@ -0,0 +1,66 @@
...
+    ConfigClass = IsrDetectorExposureMetricsAnalysisConfig
+    _DefaultName = "isrAmpOffsetMetricsAnalysis"
```
natelust, line 51: If you change above, change here

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
```diff
@@ -0,0 +1,66 @@
...
+    _SubTaskName = "ampOffset"
```
natelust, line 52: What is SubTaskName here for?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        taskName = inputRefs.metadata.datasetType.name
+        taskName = taskName[: taskName.find("_")]
+        metadata = inputs["metadata"].metadata[f"{taskName}:{self._SubTaskName}"].to_dict()
```
natelust, line 58: I see you have used `_SubTaskName` here, but why? I don't understand why this needs to be a class level variable

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
```diff
@@ -0,0 +1,66 @@
...
+        if not metadata:
+            raise NoWorkFound(f"No metadata entries for {taskName}:{self._SubTaskName}.")
+        inputs.pop("metadata")
+        df = pd.DataFrame(metadata)
+
+        inputs["data"] = df
+        outputs = self.run(**inputs)
```
natelust, line 65: You may want to proactively generate plot info like the base class does as well, so that if someone chooses to add plots or other things they can.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
natelust: I think this task is something about metadata, not about `AmpOffset`. This task could work on any other data in the metric as well, and could be configured to do that by just adding more Analysis tools to run.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
natelust: I think this is not the right place for this. Tasks should be specialized to their input dataset type, which in this case is isr metadata, not for the values inside the dataset, that is something for the analysis tools to care about. I think you need to orthogonalize these concerns

natelust (dismissed):

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
leeskelvin, line 1: The name of this file, and the classes therein, are incongruent with how many other atools classes/files are named. E.g.:

| file name                        | task name                         | default `outputName` | data input name          | data input dimensions |
|----------------------------------|-----------------------------------|----------------------|--------------------------|-----------------------|
| `objectTableTractAnalysis.py`    | `ObjectTableTractAnalysisTask`    | `objectTable_tract`  | `objectTable_tract`      | `skymap`, `tract`     |
| `sourceTableVisitAnalysis.py`    | `SourceTableVisitAnalysisTask`    |                      | `sourceTable_visit`      | `visit`, `band`       |
| `deltaSkyCorrAnalysis.py`        | `DeltaSkyCorrAnalysisTask`        | `deltaSkyCorr`       | `delta_skyCorr_hist`     | `instrument`, `visit` |

I know we don't have the strictest of naming standards inside atools, but I'd probably recommend renaming everything on this ticket to bring it in-line with our gold-standard `objectTableTractAnalysis` tooling.

For this ticket, which has `isr_metadata` as an input with dimensions `exposure`/`detector`, that would therefore imply:
* `isrMetadataExposureDetectorAnalysis.py`
* `IsrMetadataExposureDetectorAnalysisTask`
* `isr_metadata_exposure_detector`

pipelines/exposureQualityCore.yaml
```diff
@@ -0,0 +1,13 @@
...
+  Tier 1 plots and metrics to assess exposure quality
+tasks:
+  analyzeAmpOffsetTaskCore:
+    class: lsst.analysis.tools.tasks.IsrAmpOffsetMetadataAnalysisTask
+    config:
+      connections.outputName: isrAmpOffsetMetadata
+      atools.isrAmpOffsetMetadataMetric: MetadataMetricTool
+      atools.isrAmpOffsetMetadataMetric.metrics:
+        # Format is "metric name in the metadata": units
+        AMPOFFSET_PEDESTALS: adu
```
leeskelvin, line 11: Is there any merit in storing amp-offsets in nanoJanskys, rather than ADUs?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: Changing `Metrics` -> `Metadata` does not resolve the second naming concern raised by @natelust, namely, this should have amp-offset-anything in the name either. I have naming suggestions in my review below.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: Have you tested the above?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
```diff
@@ -0,0 +1,68 @@
...
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        dataId = butlerQC.quantum.dataId
+        inputs = butlerQC.get(inputRefs)
+        taskName = inputRefs.metadata.datasetType.name
+        taskName = taskName[: taskName.find("_")]
+        subTaskName = "ampOffset"
+        metadata = inputs["metadata"].metadata[f"{taskName}:{subTaskName}"].to_dict()
+        if not metadata:
+            raise NoWorkFound(f"No metadata entries for {taskName}:{subTaskName}.")
+        inputs.pop("metadata")
```
leeskelvin, line 62: Can this be simplified, combining the metadata getter and pop into a single line?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
```diff
@@ -0,0 +1,68 @@
...
+class IsrAmpOffsetMetadataAnalysisTask(AnalysisPipelineTask):
+    ConfigClass = IsrDetectorExposureMetadataAnalysisConfig
+    _DefaultName = "isrAmpOffsetMetadataAnalysis"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        dataId = butlerQC.quantum.dataId
+        inputs = butlerQC.get(inputRefs)
+        taskName = inputRefs.metadata.datasetType.name
+        taskName = taskName[: taskName.find("_")]
+        subTaskName = "ampOffset"
```
leeskelvin, line 58: Is it wise to hard-code the `subTaskName` here? What if someone wants to work with some other `isr_metadata` subtask output?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
```diff
@@ -0,0 +1,68 @@
...
+        metadata = inputs["metadata"].metadata[f"{taskName}:{subTaskName}"].to_dict()
+        if not metadata:
+            raise NoWorkFound(f"No metadata entries for {taskName}:{subTaskName}.")
+        inputs.pop("metadata")
+        df = pd.DataFrame(metadata)
```
leeskelvin, line 63: Do we really need Pandas and a dataframe here, with its only use to `**` expand it back out in the run method? Will a simple dict suffice?

leeskelvin (commented):
The comments here are probably enough to chew over as a first-round of comments. Happy to take another look after you've had a chance to implement them.

python/lsst/analysis/tools/tasks/__init__.py
```diff
@@ -13,6 +13,7 @@
 from .diffimTaskDetectorVisitSpatiallySampledAnalysis import *
 from .diffMatchedAnalysis import *
 from .gatherResourceUsage import *
+from .isrTaskDetectorExposureMetricsAnalysis import *
```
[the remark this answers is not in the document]
leeskelvin: Edit: I see you modified this between my starting the review and completing it, thanks! Don't forget to edit this again following any further name changes (comment below).

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
enourbakhsh: I tried to use `inputs["data"] = metadata` but it gives me an error:
```python
  File "/home/en7908/repos/analysis_tools/python/lsst/analysis/tools/actions/scalar/scalarActions.py", line 112, in __call__
    return cast(Scalar, float(data[self.vectorKey.format(**kwargs)][0]))
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
KeyError: 0
```
Ian also had to use a dataframe for a similar situation.

pipelines/exposureQualityCore.yaml
enourbakhsh: Hmm, do we care about cross-instrument comparisons?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
enourbakhsh: I found a way to avoid using Pandas dataframe with a custom `AnalysisAction`. Need to clean up before pushing.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
enourbakhsh: Needed modifications. Now tested.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetadataAnalysis.py
enourbakhsh: Done!

pipelines/exposureQualityCore.yaml
leeskelvin: Maybe? To put this another way, if I did want to convert these per-amp pedestals to nJy after-the-fact, how easy would that be for an end-user?

enourbakhsh: It'd take some effort - not super easy, but I double-checked with @TallJimbo, and he believes `ADU` is fine because people would be mostly interested in it within the instrument.

python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -0,0 +1,54 @@
...
+
+from typing import Any, Dict, Iterable
+
+from lsst.pex.config import DictField, Field
+
+from ..interfaces import AnalysisAction, AnalysisTool
+
+
+class LoadDoubleKeyedData(AnalysisAction):
+    """Load double keyed data"""
```
leeskelvin, line 33: Can you be more explicit here?

leeskelvin: Furthermore, this probably needs to be moved to somewhere under the `action` namespace.

python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -0,0 +1,54 @@
...
+
+    name = Field[str](doc="The name of the key to load.")
+
+    def getInputSchema(self) -> Iterable[tuple[str, Dict]]:
```
leeskelvin, line 37: Why doesn't this return a `KeyedDataSchema`? That may necessitate a modification to `KeyedDataSchema`, but it seems in-keeping with the MO elsewhere in atools, right?

python/lsst/analysis/tools/interfaces/_stages.py
```diff
@@ -223,7 +223,12 @@ def __call__(self, data: KeyedData, **kwargs) -> MetricResultType:
             if newName := self.newNames.get(key):
                 formattedKey = newName.format(**kwargs)
             notes = {"metric_tags": kwargs.get("metric_tags", [])}
-            results[formattedKey] = Measurement(formattedKey, value * apu.Unit(unit), notes=notes)
+            if isinstance(value, dict):
```
leeskelvin, line 226: I'm tempted to suggest making this more generic, and checking `if isinstance(value, Mapping):`. Is that something that might be useful down the line?

pipelines/exposureQualityCore.yaml
leeskelvin: Task renamed to `IsrMetadataExposureDetectorAnalysisTask` following review.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: Task renamed to `IsrMetadataExposureDetectorAnalysisTask` following review.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: `subTaskName` is now a config `Field` within `IsrMetadataExposureDetectorAnalysisConfig`. Does this satisfy the concerns of @natelust? I note that not supplying a `subTaskName` will default to `None`, which would allow an analysis tool to handle it.

enourbakhsh: I believe it does to a considerable extent.

python/lsst/analysis/tools/interfaces/_stages.py
enourbakhsh: It's probably best not to limit ourselves to dictionaries. I tested it, and it works.

python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -0,0 +1,54 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("MetadataMetricTool",)
+
+from typing import Any, Dict, Iterable
```
timj, line 25: Dict is not needed. In modern python you can annotate with dict directly.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Because it's not included [here](https://github.com/lsst/analysis_tools/blob/cdf17bb5f2d4739b48d7302d31eab345eaa8c960/python/lsst/analysis/tools/interfaces/_interfaces.py#L117):
```python
KeyedDataSchema = Iterable[tuple[str, type[Vector] | ScalarType | type[HealSparseMap] | type[Tensor]]]
```
I wasn't brave enough to modify `KeyedDataSchema`.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: I've made the docstring more specific. The input data is something like:
```python
{'AMPOFFSET_PEDESTALS': {'0': -0.20423207279935987, '1': -0.06727251761094916, '2': -0.28773895061785293, '3': 0.5592435410281624}}
```

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Yes, removed it from the imports. Ended up using `Mapping`.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Not sure where would be the best place to move it to, maybe above [this class](https://github.com/lsst/analysis_tools/blob/cdf17bb5f2d4739b48d7302d31eab345eaa8c960/python/lsst/analysis/tools/actions/vector/vectorActions.py#L71)?

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: Task renamed to `IsrMetadataExposureDetectorAnalysisTask` following review.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
leeskelvin: Do we need this config field at all? I.e., what's the benefit of selecting a subtask here, rather than passing the entire metadata through to an analysis tool to subselect on itself?

python/lsst/analysis/tools/atools/metadataMetrics.py
leeskelvin: [I'd imagine somewhere in here would be optimal](https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/actions/keyedData/keyedDataActions.py)? You can probably change it to inherit from `KeyedDataAction` instead as well (please test).

python/lsst/analysis/tools/atools/metadataMetrics.py
leeskelvin: Can you not squeeze your outputs into one of these types? I.e., you're currently returning a `Iterable[tuple[str, dict]]` - why does this need to be a `dict`? All you're really interested in is the amp offset value, right? Your concern is that it might make it difficult to interpret across cameras?

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: I was unsure about `KeyedDataAction` at first because I thought it had to change the data, not just load it. On second thought, I think it might be the right place. Testing it now.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Now updated to:
```python
    def getInputSchema(self) -> Iterable[tuple[str, Mapping[str, Any]]]:
        return [(self.name, Mapping[str, Any])]
```

> Can you not squeeze your outputs into one of these types? I.e., you're currently returning a `Iterable[tuple[str, dict]]` - why does this need to be a `dict`? All you're really interested in is the amp offset value, right?

I can, but then we reach a point where the pipeline yaml needs to be camera-specific, providing explicit information about the metadata for the specific camera we're interested in — not ideal. Furthermore, to squeeze our outputs into one of those types in `KeyedDataSchema`, we'll need to modify the `runQuantum` method in `IsrMetadataExposureDetectorAnalysisTask`. However, making these changes would narrow the task down to the `ampOffset` subtask and defeat the purpose of keeping it versatile for all types of ISR metadata.

> Your concern is that it might make it difficult to interpret across cameras?

Correct.

python/lsst/analysis/tools/tasks/isrTaskDetectorExposureMetricsAnalysis.py
enourbakhsh: Agreed, we should be able to pass the entire metadata to an analysis tool for subselection. We just need a more sophisticated action to locate the data. I'll test it out.

enourbakhsh: Done! Needed some refactoring.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Made a more generic action named `KeyedDataKeyAccessAction` that inherits from `KeyedDataAction` and moved it to `keyedDataActions.py`.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: Following your suggestion, I slightly modified `KeyedData`, `KeyedDataSchema`, and `KeyedDataTypes` to include the `Mapping` type and stay aligned with the MO elsewhere in atools.

python/lsst/analysis/tools/interfaces/_interfaces.py
```diff
@@ -103,18 +103,23 @@ def __init__(self) -> None:
 like an NDArray should be considered a Vector.
 """
 
-KeyedData = MutableMapping[str, Vector | Scalar | HealSparseMap | Tensor]
-"""KeyedData is an interface where either a `Vector` or `Scalar` can be
-retrieved using a key which is of str type.
+KeyedData = MutableMapping[str, Vector | Scalar | HealSparseMap | Tensor | Mapping[str, Any]]
```
leeskelvin, line 106: Does `Mapping` need `[str, Any]` here?

leeskelvin: (and elsewhere below)

python/lsst/analysis/tools/interfaces/_stages.py
```diff
@@ -218,12 +218,19 @@ def __call__(self, data: KeyedData, **kwargs) -> MetricResultType:
             if formattedKey not in data:
                 raise ValueError(f"Key: {formattedKey} could not be found input data")
             value = data[formattedKey]
-            if not isinstance(value, Scalar):
-                raise ValueError(f"Data for key {key} is not a Scalar type")
```
leeskelvin, line 222: Curious why you removed this `raise`, rather than add `Mapping` as an exception and keep it?

leeskelvin: The `raise` below will only trigger now if `isinstance(value, Mapping)` is `True`, which doesn't capture the same functionality as this prior `raise`.

leeskelvin: I see it in the `else` block now - I suppose that achieves the same thing, but I did prefer failing sooner rather than later.

leeskelvin (approved):

python/lsst/analysis/tools/interfaces/_stages.py
enourbakhsh: Correct, I just moved it elsewhere. Also, my tests show that `isinstance(value, Scalar)` doesn't work as expected:
```python
>>> from lsst.analysis.tools.interfaces import Scalar
>>> isinstance("string", Scalar)
True
>>> isinstance(["string1","string2"], Scalar)
True
>>> isinstance([1, 2, 3], Scalar)
True
>>> isinstance(5.5, Scalar)
True
>>> isinstance([5.5, 8.7], Scalar)
True
>>> isinstance({}, Scalar)
True
>>> isinstance({"key1": 2.60, "key2": 9.13}, Scalar)
True
```

python/lsst/analysis/tools/interfaces/_interfaces.py
enourbakhsh: I was under the impression that `Mapping` needs to specify the key and value types. I can double-check.

python/lsst/analysis/tools/interfaces/_stages.py
timj: @natelust do you expect this? The code for `Scalar` seems to imply that it's meant to only be numbers and not any other type.

leeskelvin: @natelust is on vacation at present, and is unlikely to respond before this ticket is merged. My suggestion is for this to be flagged on a new ticket, and addressed once @natelust gets back.

enourbakhsh: I just showed this to @sr525, and she thinks we should file a ticket for it and have @natelust  look at it. I'll file one.

python/lsst/analysis/tools/interfaces/_interfaces.py
enourbakhsh: Tested it without using `[str, Any]` and there were no issues.

python/lsst/analysis/tools/interfaces/_stages.py
enourbakhsh: Filed [DM-45309](https://rubinobs.atlassian.net/browse/DM-45309) and added @natelust as a watcher.

PR #268, opened 2024-06-17
DM-43849: Create spatiallySampledMetric to visualize the diffim kernel

BrunoSanchez:
Added a quiver-plot analysis tool and use it for Ap in the `apDetectorVisitQualityCore.yaml`.


pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -83,6 +83,14 @@ tasks:
       connections.outputName: diffimPlots
       connections.fakesType: parameters.fakesType
       connections.coaddName: parameters.coaddName
+      atools.dipoleQuiver: DiffimSpatialMetricsDipoleQuiverPlot
+      atools.dipoleQuiver.angleName: dipole_direction
+      atools.dipoleQuiver.lengthName: dipole_separation
+      atools.dipoleQuiver.produce.plot.qKeyLabel: "0.2 arcsec Avg Dipole length"
```
isullivan, line 89: Is this right my read of `ax.quiverkey(q, 0.5, 0.9, 0.1, self.qKeyLabel, labelpos="E", coordinates="figure")` is that the quiver label is 0.1 arcsec.

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -83,6 +83,14 @@ tasks:
...
+      atools.kernelQuiver: DiffimSpatialMetricsDipoleQuiverPlot
+      atools.kernelQuiver.angleName: psfMatchingKernel_position_angle
+      atools.kernelQuiver.lengthName: psfMatchingKernel_length
+      atools.kernelQuiver.produce.plot.qKeyLabel: "0.2 arcsec Kernel offset direction"
```
isullivan, line 93: Also check this 0.2 or 0.1 arcsec

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -83,6 +83,14 @@ tasks:
       connections.outputName: diffimPlots
       connections.fakesType: parameters.fakesType
       connections.coaddName: parameters.coaddName
+      atools.dipoleQuiver: DiffimSpatialMetricsDipoleQuiverPlot
+      atools.dipoleQuiver.angleName: dipole_direction
+      atools.dipoleQuiver.lengthName: dipole_separation
+      atools.dipoleQuiver.produce.plot.qKeyLabel: "0.2 arcsec Avg Dipole length"
+      atools.kernelQuiver: DiffimSpatialMetricsDipoleQuiverPlot
+      atools.kernelQuiver.angleName: psfMatchingKernel_position_angle
```
isullivan, line 91: After thinking about this longer, I agree with you that this would make more sense in detector coordinates using `psfMatchingKernel_direction`, even if that is different from the dipole plots.
Ideally we would also have a marker indicating the direction towards zenith, but that seems beyond the scope of the current ticket.

isullivan (approved):
Since the quiver plot is no longer just for dipoles, I suggest removing `dipole` from the class name `DiffimSpatialMetricsDipoleQuiverPlot`.
I also agree that the kernel quiver plot should be based on the detector angle, not position angle, so I suggest you switch that back.

pipelines/apDetectorVisitQualityCore.yaml
BrunoSanchez: Now we include another parameter to set up this size accordingly.

pipelines/apDetectorVisitQualityCore.yaml
BrunoSanchez: This is corrected now. Thanks.

PR #270, opened 2024-06-18
DM-44891: Fix bug when zero-length data is passed to calculateRange.py/MinMax

jrmullaney:
Prevents a catastrophic fail should a zero-length vector be passed to `MinMax`.


mrawls (approved):
Looks good, but I think it might be easiest to incorporate this small change into DM-44869. If that's not actually easier it's also fine for this to be its own tiny ticket.

PR #271, opened 2024-06-18
DM-44854: Add expTime and pixelScale to visitSummary and ccdVisitTable

laurenam:


jrmullaney (approved):
Frustrating that we can't use compound units. I wonder whether this can be added (obviously way beyond the scope of this ticket - I'm not suggesting you add that feature on this ticket!!).

PR #274, opened 2024-07-12
DM-45226: Modify nan/inf usage to support numpy 2.0

timj:


python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -517,7 +517,7 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
             case _:
                 bands = None
         if bands:
-            mask = np.in1d(data[self.vectorKey], bands)
+            mask = np.isin(data[self.vectorKey], bands)
```
timj, line 520: @bechtol you wrote this line originally. Does this change look okay to you? I think the difference between the two is that the former always returned a 1-D array but now it returns something in the same shape as the original.

kherner: I'm sure you already read it but to post for the record, in1d is officially deprecated in 2.0+ and isin is the replacement. Is there any case where the shape preservation of isin() would be a problem?

[numpy documentation on in1d](https://numpy.org/doc/stable/reference/generated/numpy.in1d.html)

kherner: Personally I can't think of anything but I am far from 100% knowledgeable on such things.

kherner (approved):
See comments.

python/lsst/analysis/tools/actions/vector/selectors.py
timj: The issue is if `mask` is expected to be 1D but `data` is N-D, in which case the return from `isin` presumably has to be turned into 1-D. The mask created by `np.full` is going to be 1D.

timj: Unfortunately this method is never called in tests.

timj: Ok. I'm going to change this to call `.flatten()` to guarantee it's returning the same thing.

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -517,7 +517,7 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
             case _:
                 bands = None
         if bands:
-            mask = np.in1d(data[self.vectorKey], bands)
+            mask = np.isin(data[self.vectorKey], bands).flatten()
```
kherner, line 520: Fair enough.

PR #277, opened 2024-08-02
DM-34048: Add completeness/purity plots

taranu:


pipelines/coaddDiffMatchedInjectedCore.yaml
```diff
@@ -0,0 +1,201 @@
+description: |
+  Matched difference (measured vs reference) plots/metrics
+parameters:
```
sr525, line 3: I think more of this should be in the atool and less in the pipeline file.

pipelines/coaddDiffMatchedInjectedCore.yaml
```diff
@@ -0,0 +1,201 @@
...
+      atools.matchedRefPsfMagDiffZoom.produce.plot.yLims: lims_mag_diff_zoom
+      atools.matchedRefPsfMagDiffZoom.produce.metric: NoMetric
+
+      atools.matchedRefPsfMagChi: MatchedRefCoaddDiffMagTool
+      atools.matchedRefPsfMagChi.compute_chi: true
+      atools.matchedRefPsfMagChi.produce.plot.xLims: lims_mag_x_full
+      atools.matchedRefPsfMagChi.produce.plot.yLims: lims_mag_chi
+
+#      atools.matchedRefPositionXDiff: MatchedRefCoaddDiffPositionTool
+#      atools.matchedRefPositionXDiff.coord_meas: x
```
sr525, line 84: Is this meant to all still be in here?

pipelines/coaddDiffMatchedInjectedCore.yaml
```diff
@@ -0,0 +1,201 @@
...
+      atools.matchedRefPositionDecDiffZoom.produce.metric: NoMetric
+
+      atools.matchedRefPositionDecChi: MatchedRefCoaddDiffPositionTool
+      atools.matchedRefPositionDecChi.coord_meas: coord_dec
+      atools.matchedRefPositionDecChi.coord_ref: ref_dec
+      atools.matchedRefPositionDecChi.produce.plot.xLims: lims_mag_x_full
+      atools.matchedRefPositionDecChi.produce.plot.yLims: lims_pos_chi
+      atools.matchedRefPositionDecChi.compute_chi: true
+
+      python: |
```
sr525, line 152: I also don't like all of this being in here but I think that is personal preference. I think it would look tidier if more of this was in the atool and less in the pipeline.

python/lsst/analysis/tools/actions/config/binning.py
```diff
@@ -0,0 +1,64 @@
...
+
+class MagnitudeBinnedMetricsConfig(pexConfig.Config):
+    """Configuration for magnitude-binned metrics.
+
+    Bins may only be specified in mmag widths
+    """
+
+    mag_low_min = pexConfig.Field[int](
+        doc="Lower bound for the smallest magnitude bin in millimags",
+        default=15000,
```
sr525, line 36: This being in millimags seems silly.

python/lsst/analysis/tools/actions/config/binning.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    def get_bins(self):
+        return list(range(self.mag_low_min, self.mag_low_max, self.mag_interval))
+
+    def get_name_bin(self, value_bin: int) -> str:
+        prefix = str(value_bin // 1000)
+        suffix = str(value_bin % 1000)
+        name = f"{prefix}{'p' if suffix else ''}{suffix}"
+        return name
```
sr525, line 60: Can we have some sort of doc strings for these? Particularly get_name_in is not obvious to me.

python/lsst/analysis/tools/actions/keyedData/calcBinnedCompleteness.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+    key_mask_target = Field[str](
+        default=None,
+        doc="Key for mask to apply for target objects in purity",
+        optional=True,
+    )
+    name_prefix = Field[str](default="", doc="Field name to append statistic names to")
+    name_suffix = Field[str](default="", doc="Field name to append to statistic names")
+    selector_range_ref = ConfigurableActionField[RangeSelector](doc="Range selector")
+    selector_range_target = ConfigurableActionField[RangeSelector](doc="Range selector")
```
sr525, line 56: These should have different docs.

python/lsst/analysis/tools/actions/keyedData/calcBinnedCompleteness.py
```diff
@@ -0,0 +1,227 @@
...
+import numpy as np
+from lsst.pex.config import Field
+from lsst.pex.config.configurableActions import ConfigurableActionField
+
+from ...interfaces import KeyedData, KeyedDataAction, KeyedDataSchema, Scalar, Vector
+from ...math import divide
+from ..vector.selectors import RangeSelector
+
+
+class CalcBinnedCompletenessAction(KeyedDataAction):
```
sr525, line 34: Somewhere this needs a decent doc string that tells you what it does and what it returns.

python/lsst/analysis/tools/actions/keyedData/calcCompletenessHistogram.py
```diff
@@ -0,0 +1,127 @@
...
+        doc="The action to compute completeness/purity",
+    )
+    bins = pexConfig.ConfigField[MagnitudeBinnedMetricsConfig](
+        doc="The bin configuration",
+    )
+    config_metrics = pexConfig.ConfigField[MagnitudeCompletenessConfig](doc="Metric definition configuration")
+
+    def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
+        band = kwargs.get("band")
+        bins = tuple(x / 1000.0 for x in reversed(self.bins.get_bins()))
```
sr525, line 65: Should the number of bins be a config option?

python/lsst/analysis/tools/actions/keyedData/calcCompletenessHistogram.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class CalcCompletenessHistogramAction(KeyedDataAction):
+    """Action to calculate a histogram of completeness vs magnitude."""
+
+    action = ConfigurableActionField[CalcBinnedCompletenessAction](
+        doc="The action to compute completeness/purity",
+    )
+    bins = pexConfig.ConfigField[MagnitudeBinnedMetricsConfig](
+        doc="The bin configuration",
```
sr525, line 59: What does "The bin configuration" mean?

python/lsst/analysis/tools/actions/plot/completenessPlot.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns
+        ------
+        ``fig``
+            The figure to be saved (`matplotlib.figure.Figure`).
+
+        Notes
+        -----
+        Makes a histogram showing the fraction recovered in each magnitude
+        bin with the number input and recovered overplotted.
+        """
```
sr525, line 111: Please include an example plot here, there is an example of how to do that in scatterPlotWithTwoHists.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -444,3 +446,39 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+            values,
+            range=(np.nanmin(values), np.nanmax(values)),
+            bins=100,
+        )
+        nOutput, _ = np.histogram(
+            values[matched],
+            range=(np.nanmin(values[matched]), np.nanmax(values[matched])),
+            bins=bins,
+        )
+        # Find bin where the fraction recovered first falls below 0.5
```
sr525, line 478: I think this comment is left over from before.

python/lsst/analysis/tools/actions/vector/calcBinnedCompleteness.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class CalcBinnedCompletenessAction(KeyedDataAction):
+    name_prefix = Field[str](default="", doc="Field name to append statistic names to")
+    name_suffix = Field[str](default="", doc="Field name to append to statistic names")
+    selector_type_match = ConfigurableActionField[VectorAction](
+        doc="Selector to return objects with matching ref and target type classification"
+    )
+    selector_range_ref = ConfigurableActionField[RangeSelector](doc="Range selector")
+    selector_range_target = ConfigurableActionField[RangeSelector](doc="Range selector")
```
sr525, line 42: These need different doc strings.

python/lsst/analysis/tools/actions/vector/calcBinnedCompleteness.py
```diff
@@ -0,0 +1,145 @@
+# This file is part of analysis_tools.
```
sr525, line 1: How is this different to the one in keyedData?

taranu: It is not; I was supposed to have deleted this one since it returns KeyedData and should not be under vector.

python/lsst/analysis/tools/atools/diffMatched.py
```diff
@@ -115,86 +134,260 @@ def setDefaults(self):
         self.vectorKey = "refcat_is_pointsource"
 
 
-class MatchedRefCoaddToolBase(MagnitudeXTool):
+class InjectedObjectSelector(SelectorBase):
+    """A selector for injected objects."""
```
sr525, line 138: Why is this selector being defined in the atool and not in vector actions?

python/lsst/analysis/tools/atools/diffMatched.py
```diff
@@ -115,86 +134,260 @@ def setDefaults(self):
...
+            is_target_galaxy = is_target_galaxy[mask]
+            is_target_star = is_target_star[mask]
+        return (is_ref_galaxy & is_target_galaxy) | (is_ref_star & is_target_star)
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        yield self.key_is_ref_galaxy, Vector
+        yield self.key_is_ref_star, Vector
+        yield self.key_is_target_galaxy, Vector
+        yield self.key_is_target_star, Vector
+
```
sr525, line 230: All of these actions should be defined in the appropriate actions file. Or make a new one for injected stuff but other people may want to use these things and they should be in the actions folder where people will find them.

python/lsst/analysis/tools/atools/diffMatched.py
```diff
@@ -231,25 +422,31 @@ def configureMetrics(
...
-                action = getattr(self.process.calculateActions, f"{name}{minimum}{attr_suffix}")
+            # Set up metrics for objects of one class within a magnitude range
+            bins = self.mag_bins.get_bins()
+            for minimum in bins:
+                action = getattr(self.process.calculateActions, f"{name_type_plural}_{minimum}{attr_suffix}")
                 action.selector_range = RangeSelector(
                     vectorKey=x_key,
-                    minimum=minimum,
-                    maximum=minimum + self._mag_interval,
+                    minimum=minimum / 1000.0,
```
sr525, line 438: It seems particularly silly for these to be defined in millimags and then /1000 everywhere.

sr525:
A general comment I have is that atools should be fairly human readable and easy to understand, someone unfamiliar with the code should be able to look at it and see what is going on. This is particularly true for plots with injected sources where science users will, hopefully, see Rubin diagnostic plots and want to make them for their own injected data. These atools are structured differently from most others and fairly opaque. Is there anyway to make the atools more human readable to a newer user?

python/lsst/analysis/tools/atools/genericBuild.py
```diff
@@ -106,9 +116,139 @@ class FluxesDefaultConfig(Config):
     ref_matched = ConfigField[FluxConfig](doc="Reference catalog magnitude")
```
sr525, line 118: Why are these things in genericBuild?

pipelines/coaddDiffMatchedInjectedCore.yaml
taranu: It would need the x/y columns to be in the matched table to work. I'm debating whether to add them and make both x/y and ra/dec plots. Perhaps @cmsaunders has thoughts? I figure the difference plots will be uninteresting if the WCS is good (and if not, then it will show up elsewhere), but the chi plots might be a worthwhile sanity check for ra/dec errors.

pipelines/coaddDiffMatchedInjectedCore.yaml
taranu: These are here to make it easier to change settings for all of the relevant plots at once, which is something I thought users would be reasonably likely to want to do.

On the other hand, I don't really like pipeline parameters either. Doing it this way means they have to be the same for all of the plots; if anyone wants to change a few here or there, they'd have to copy the file and change them per-tool anyway. There's no easy way I can think of to keep these in sync with the actual config defaults either.

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
```diff
@@ -0,0 +1,286 @@
...
+        self.produce.metric.newNames = {"mag20": "{band}_mag20", "mag50": "{band}_mag50"}
+
+
+class TargetInjectedCatDelta(AnalysisTool):
+    """Plot the difference between a target catalog and an
+    injected catalog for the quantity set in `setDefaults`.
+    """
+
+    parameterizedBand = Field[bool](
+        doc="Does this AnalysisTool support band as a name parameter", default=True
```
sr525, line 83: This doesn't need to be here, just set parametrizeBand = True/False at the top.

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
```diff
@@ -0,0 +1,286 @@
...
+    of a target catalog and an injected catalog.
+    """
+
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.process.buildActions.yStars.col1 = "{band}_cModelFlux"
+
+
+# TODO: get this to actually work for photometric metrics
```
sr525, line 200: Is this TODO still valid?

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
```diff
@@ -0,0 +1,286 @@
...
+class TargetInjectedCatDeltaMetrics(AnalysisTool):
+    """Calculate the diff metric and the sigma MAD from the difference between
+    the target and injected catalog coordinates and photometry.
+    """
+
+    parameterizedBand = Field[bool](
+        doc="Does this AnalysisTool support band as a name parameter", default=True
```
sr525, line 207: Same as above

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+
+    def coaddContext(self) -> None:
+        """Apply coadd options for the metrics. Applies the coadd plot flag
+        selector and sets flux types.
+        """
+
+        # self.applyContext(RefMatchContext)
```
sr525, line 215: Comment

python/lsst/analysis/tools/tasks/injectedObjectAnalysis.py
```diff
@@ -0,0 +1,64 @@
...
+from lsst.skymap import BaseSkyMap
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class InjectedObjectAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions=("skymap", "tract"),
+    defaultTemplates={
+        "outputName": "matched_injected_deepCoadd_catalog_tract_injected_objectTable_tract",
```
sr525, line 35: Does injected need to be in this name twice? I am not sure where it is actually named and guess that it is not on this ticket.

python/lsst/analysis/tools/tasks/injectedObjectAnalysis.py
```diff
@@ -0,0 +1,64 @@
...
+
+__all__ = ("InjectedObjectAnalysisConfig", "InjectedObjectAnalysisTask")
+
+from lsst.pipe.base import connectionTypes as ct
+from lsst.skymap import BaseSkyMap
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class InjectedObjectAnalysisConnections(
```
sr525, line 31: I do think that consolidating this with objectTableTractAnalysis would be better unless there is a compelling reason not to?

python/lsst/analysis/tools/tasks/objectTableTractAnalysis.py
```diff
@@ -34,11 +36,11 @@
 class ObjectTableTractAnalysisConnections(
     AnalysisBaseConnections,
     dimensions=("skymap", "tract"),
-    defaultTemplates={"outputName": "objectTable_tract"},
+    defaultTemplates={"inputName": "objectTable_tract", "outputName": "objectTable_tract"},
```
sr525, line 39: Thank you for generalising this.

tests/test_completenessPlot.py
```diff
@@ -0,0 +1,183 @@
...
+import tempfile
+import unittest
+
+import lsst.utils.tests
+import matplotlib.pyplot as plt
+import numpy as np
+from lsst.analysis.tools.actions.plot import CompletenessHist
+from lsst.analysis.tools.actions.plot.plotUtils import get_and_remove_figure_text
+
+# matplotlib.use("Agg")
```
sr525, line 34: Comment

pipelines/coaddDiffMatchedInjectedCore.yaml
sr525: I can understand leaving it for later if it is going to be added back in but then maybe add a note and a TODO to say why it is there.

python/lsst/analysis/tools/actions/config/binning.py
sr525: Does going from 15 to 30 really make the maths than much harder than going from 15000 to 30000? It is much more understandable to go from 15 to 30 at a glance than 15000.

pipelines/coaddDiffMatchedInjectedCore.yaml
cmsaunders: I didn't look through the whole ticket to fully get the context, but if this is going to show ra/dec for the pipeline position versus the injected position, that sounds interesting to me.

taranu: There will be ra/dec, the question is whether to also make plots of x/y (and particularly chi=(x_meas - x_ref)/x_err).

python/lsst/analysis/tools/actions/config/binning.py
taranu: I wanted to specify a min, max and width (rather than a number of bins with the width implicit) and not have rounding remove/add one bin because it ended on 23.9999999 instead of 24.0.

python/lsst/analysis/tools/tasks/injectedObjectAnalysis.py
jtmccann: I believe this comes from [DM-41210](https://rubinobs.atlassian.net/browse/DM-41210). The format for the connection name is "matched_{catalog1}_{catalog2}". Here both catalogs being matched happen to have "injected_" in the name.

python/lsst/analysis/tools/actions/keyedData/calcCompletenessHistogram.py
taranu: The bins are returned by a config class.

python/lsst/analysis/tools/actions/keyedData/calcCompletenessHistogram.py
taranu: It is a config field. Having said that, I am renaming it to `MagnitudeBinConfig` because it doesn't define anything specific to metrics.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
taranu: The whole action is.

python/lsst/analysis/tools/tasks/injectedObjectAnalysis.py
taranu: I also lean towards consolidation, but if they ever needed to diverge in the future, we'd have to add the class in again, editing every pipeline where it's used. Otherwise, the only benefit is having different defaults. 

In retrospect we probably should have called `ObjectTableTractAnalysis` `TractTableAnalysis` since it's only the defaults that point to objectTable_tract.

python/lsst/analysis/tools/actions/vector/calcBinnedCompleteness.py
taranu: Changed in the one I didn't delete.

python/lsst/analysis/tools/atools/diffMatched.py
taranu: No particularly compelling reason.

python/lsst/analysis/tools/atools/diffMatched.py
taranu: `actions/vector/selectors.py` is starting to get rather long. Any name suggestions? `injectedSelectors.py` is more readable but `selectorsInjected.py` sorts better.

python/lsst/analysis/tools/tasks/injectedObjectAnalysis.py
taranu: That's the default connection name format for matched tables. Attempting to shorten the name would just make it harder to figure out which tables were matched.

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
taranu: These are all Trys' original plots using the atools matcher that are no longer being used, sorry for making you read them (again).

tests/test_completenessPlot.py
taranu: Actually this needs to be there or else pytesting the file will actually open a plot window, although weirdly that doesn't happen with scons so scons must also be setting that somewhere.

python/lsst/analysis/tools/atools/diffMatched.py
jtmccann: Both are readable to me, but maybe I'm biased by familiarity.

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
jtmccann: Are any of my plots getting used? I have them all saved on another PR so you can delete them if they're redundant.

pipelines/coaddDiffMatchedInjectedCore.yaml
taranu: I did what I could to simplify this by consolidating necessary overrides into a single function call in the python block. That function will almost almost always need to be called by anything overriding the task - in obs_subaru, for example, it needs to set the bands for the colour diff plots. I might consider changing the default context from `dc2` to `injection`.

pipelines/coaddDiffMatchedInjectedCore.yaml
taranu: See the note below. I dropped the parameters entirely.

I could re-implement the ability to mass change limits etc. in `reconfigure_diff_matched_defaults` if needed.

This file also ends up doing the same thing as the original DC2 `/coaddDiffMatchedQualityExtended.yaml` pipeline, which I have not updated here. They could be consolidated at some point, or have one base that they each import.

python/lsst/analysis/tools/atools/diffMatched.py
taranu: We do tend use mmag on the y axis.

python/lsst/analysis/tools/atools/genericBuild.py
taranu: The selectors have been moved to the selectors file.

python/lsst/analysis/tools/atools/sourceInjectionPlots.py
taranu: Not here, no. I only kept what is still being used from your original commits.

tests/test_completenessPlot.py
taranu: I added a boolean (default `False`) to disable this for debugging tests.

pipelines/coaddDiffMatchedInjectedCore.yaml
sr525: The different bands, between obs_subaru and obs_decam for example, should be set in the obs specific config files.

PR #278, opened 2024-08-02
DM-39168: Add proper motion/parallax correction to astrometry metrics

cmsaunders:


python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+        doc="Catalog of epochs for refCat objects.",
+        name="objectEpochTable",
+        storageClass="DataFrame",
+        dimensions=["skymap", "tract", "patch"],
+        multiple=True,
+    )
+
+
+class ObjectEpochTableConfig(pipeBase.PipelineTaskConfig, pipelineConnections=ObjectEpochTableConnections):
+    refCatBands = pexConfig.ListField(
```
parejkoj, line 81: Why not just `bands`, since the object table is the only input to this class.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+        doc=("Bands in refCat to be combined with `refCat_selectors` to build refCat column names."),
+        dtype=str,
+        default=["u", "g", "r", "i", "z", "y"],
+    )
+
+
+class ObjectEpochTableTask(pipeBase.PipelineTask):
+    """Collect mean epochs for the observations that went into each object."""
```
parejkoj, line 89: I'm concerned that we don't have `epoch` or `mean_epoch` or [radecMjdTai](https://sdm-schemas.lsst.io/apdb.html#DiaObject.radecMjdTai) (as is used in the AP DiaObject table) already defined in our object tables. That feels like an oversight, and not something we should be computing in an analysis tool.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+        # "associatedSourcesInputName": "isolated_star_sources",
+        # "outputName": "sourceObjectTable",
+    },
+):
+    refCat = ct.Input(
+        doc="Catalog of positions to use as reference.",
+        name="objectTable",
+        storageClass="DataFrame",
+        dimensions=["skymap", "tract", "patch"],
+        multiple=True,
```
parejkoj, line 57: Why is this `multiple`? Oh, I think I see, this catalog is per-band? Please note that in the docstring.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -85,6 +223,11 @@ class SourceObjectTableAnalysisConfig(
         dtype=str,
         default="r_dec",
     )
+    epoch_column = pexConfig.Field(
+        doc="Name of column in refCat to use for epoch.",
+        dtype=str,
+        default="r_epoch",
```
parejkoj, line 229: This will then always use the `r`-band epoch? That doesn't seem correct.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -98,35 +241,147 @@ class SourceObjectTableAnalysisConfig(
...
+        targetEpochs = refEpochs.to_numpy()
+        # There may not be a valid reference epoch on the edge of a given
+        # region. Do not make a proper motion correction for any sources on the
+        # edge.
+        targetEpochs[~np.isfinite(targetEpochs)] = sourceMjd
+
+        with Matcher(isolatedSources["coord_ra"], isolatedSources["coord_dec"]) as m:
+            idx, i1, i2, d = m.query_radius(
+                properMotionCatalog["starX"], properMotionCatalog["starY"], 0.2 / 3600, return_indices=True
+            )
```
parejkoj, line 296: This looks very similar to the code in `applyProperMotionCorrections`: can you consolidate some code between that and this?

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -98,35 +241,147 @@ class SourceObjectTableAnalysisConfig(
...
+        raCorrection, decCorrection = calculate_apparent_motion(pmDf.iloc[i2], targetEpochs[i1])
+        isolatedSources["coord_ra"][i1] -= raCorrection.value
+        isolatedSources["coord_dec"][i1] -= decCorrection.value
+        return isolatedSources
 
+    def prepareAssociatedSources(
+        self, visit, data, associatedSourceRefs, refCats, visitTable, properMotionCatalog
+    ):
+        """Match isolated sources with reference objects and shift the sources
+        to the object epochs if the properMotionCatalog is not None.
```
parejkoj, line 316: I don't see anything in this method that allows `properMotionCatalog` to be None?

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -98,35 +241,147 @@ class SourceObjectTableAnalysisConfig(
...
+
+        Parameters
+        ----------
+        visit : `int`
+            Identifier of the visit corresponding to the data.
+        data : `astropy.table.Table`
+            Catalog of sources.
```
parejkoj, line 323:
```suggestion
            Catalog of sources to be astrometrically corrected.
```

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -98,35 +241,147 @@ class SourceObjectTableAnalysisConfig(
...
 
         if len(isolatedSources) == 0:
-            raise NoWorkFound(f"No isolated sources found for visit {visit}")
+            raise pipeBase.NoWorkFound(f"No isolated sources found for visit {visit}")
+
+        with Matcher(isolatedSources["coord_ra"], isolatedSources["coord_dec"]) as m:
+            idx, isolatedMatchIndices, refMatchIndices, dists = m.query_radius(
+                refCats[self.config.ra_column].values,
+                refCats[self.config.dec_column].values,
+                1 / 3600.0,
```
parejkoj, line 349: Why is this 1" and the others were 0.2"? Again, a magic number should at least be a function kwarg, and probably a config field.

parejkoj (changes requested):
I'll note that this is fairly complicated code but there are no tests of it. At the very least, tests of of `applyProperMotionCorrections` and `getEpochs`, since they do a bunch of dataframe shuffling and indexing internally.

I know it wasn't introduced on this ticket, but we *have to* fix the gbdes outputs to not have `starX/starY` be ra/dec. That's incredibly misleading.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
 
 
+class ObjectEpochTableConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("tract", "skymap"),
+    defaultTemplates={
+        # "inputName": "sourceTable_visit",
+        # "inputCoaddName": "deep",
+        # "associatedSourcesInputName": "isolated_star_sources",
+        # "outputName": "sourceObjectTable",
```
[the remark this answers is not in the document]
cmsaunders: Fixed, didn't mean to include `defaultTemplates` here.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -123,6 +161,76 @@ def prepareAssociatedSources(cls, skymap, tract, sourceCatalogs, associatedSourc
...
+        dataJoined : `astropy.table.Table`
+            Table containing source positions.
+        properMotionCatalogs: `dict` [`pd.DataFrame`]
+            Dictionary keyed by band with proper motion and parallax catalogs.
+        visitTable : `pd.DataFrame`
+            Table containing the MJDS of the visits.
+
+        Returns
+        -------
+        dataJoined : `astropy.table.Table`
```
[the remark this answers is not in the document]
cmsaunders: I know I don't have to, but I think modifying input variables is generally not expected in python, so I think the behavior is more clear this way.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -70,11 +74,38 @@ class AssociatedSourcesTractAnalysisConnections(
         dimensions=("instrument",),
         isCalibration=True,
     )
+    properMotionCatalogs = ct.Input(
+        doc="Catalog containing proper motions.",
+        name="gbdesAstrometricFit_starCatalog",
```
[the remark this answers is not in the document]
cmsaunders: No, I don't think the source tables will ever have proper motion and parallax columns. Soonish there should be some sort of Calibration Objects table that will have these values, but that doesn't exist yet.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -123,6 +161,76 @@ def prepareAssociatedSources(cls, skymap, tract, sourceCatalogs, associatedSourc
...
+            ipdb.set_trace()
+            catRAs = np.zeros(len(meanRAs))
+            catDecs = np.zeros(len(meanRAs))
+            pmRAs = np.zeros(len(meanRAs))
+            pmDecs = np.zeros(len(meanRAs))
+            parallaxes = np.zeros(len(meanRAs))
+            catRAs[i1] = bandPMs["starX"][i2]
+            catDecs[i1] = bandPMs["starY"][i2]
+            pmRAs[i1] = bandPMs["starPMx"][i2]
+            pmDecs[i1] = bandPMs["starPMy"][i2]
```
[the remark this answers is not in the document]
cmsaunders: The catalog in question was not initially intended to be used by downstream tasks. Converting it to something more user-friendly is on my to-do list, and I'm going to make an actual ticket so that I can cite it here.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -123,6 +161,76 @@ def prepareAssociatedSources(cls, skymap, tract, sourceCatalogs, associatedSourc
...
+                    "obj_index": meanRAs.index,
+                }
+            )
+
+            dataWithPM = pd.merge(bandSources, pmDf, on="obj_index", how="left")
+
+            visits = bandSources["visit"].unique()
+            mjds = [visitTable.loc[visit]["expMidptMJD"] for visit in visits]
+            mjdDf = pd.DataFrame({"MJD": mjds, "visit": visits})
+            dataWithMJD = pd.merge(dataWithPM, mjdDf, on="visit", how="left")
```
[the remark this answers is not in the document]
cmsaunders: I don't follow. There are multiple merges to match the sources with their motions and epochs. I don't think `calculate_apparent_motion` should have to do that sorting.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -135,6 +243,18 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             sourceCatalogs.append(self.loadData(handle, names))
         inputs["sourceCatalogs"] = sourceCatalogs
 
+        if self.config.applyProperMotions:
+            properMotions = {}
+            for pmCatRef in inputs["properMotionCatalogs"]:
+                pmCat = pmCatRef.get(
+                    parameters={"columns": ["starX", "starY", "starPMx", "starPMy", "starParallax"]}
```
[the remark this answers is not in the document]
cmsaunders: I made these a config parameter.

The second point is going to be dealt with on DM-45845.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+)
 
 import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
 import numpy as np
 import pandas as pd
 from astropy.table import vstack
-from lsst.pipe.base import NoWorkFound
+from lsst.drp.tasks.gbdesAstrometricFit import calculate_apparent_motion
 from lsst.pipe.base import connectionTypes as ct
```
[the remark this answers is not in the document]
cmsaunders: Objection noted, but that's not on this ticket!

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+        name="deepCoadd_epoch_map_mean",
+        storageClass="HealSparseMap",
+        dimensions=("skymap", "tract", "band"),
+        multiple=True,
+        deferLoad=True,
+    )
+
+    objectEpochs = ct.Output(
+        doc="Catalog of epochs for refCat objects.",
+        name="objectEpochTable",
```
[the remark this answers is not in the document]
cmsaunders: It would be nice if the `objectTables` had the epoch, but it's not possible to add that as part of this ticket. I don't see anything about not including `table` in dataset type names, unless you extend the point about catalogs to tables, but I'm fine calling this `object_epoch`.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+            for positions in the reference catalog.
+
+        Returns
+        -------
+        epochDf = `pd.DataFrame`
+            Catalog with mean epoch of visits at each object positions.
+        """
+        epochs = {}
+        for band in self.config.refCatBands:
+            refEpochs = np.ones(len(cat)) * np.nan
```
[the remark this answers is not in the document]
cmsaunders: If `validPositions.any()` is False, you wouldn't go into the block where that happens. I wanted to do
`refEpochs[validPositions][refEpochsValid] = bandEpochs`
to avoid having a second line setting `bandEpochs[~refEpochsValid] = np.nan`, which I do currently, but this doesn't work.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -20,19 +20,127 @@
...
+            validPositions = np.isfinite(cat[f"{band}_ra"]) & np.isfinite(cat[f"{band}_dec"])
+            if validPositions.any():
+                bandEpochs = epochMapDict[band].get_values_pos(
+                    cat[f"{band}_ra"][validPositions], cat[f"{band}_dec"][validPositions]
+                )
+                refEpochsValid = epochMapDict[band].get_values_pos(
+                    cat[f"{band}_ra"][validPositions], cat[f"{band}_dec"][validPositions], valid_mask=True
+                )
```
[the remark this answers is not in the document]
cmsaunders: You can have invalid positions and also invalid epochs. There are edge cases where the object position is outside the footprint covered by the healsparse map.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: See the thread here: https://lsstc.slack.com/archives/C2JPMCF5X/p1721331152029129. It might go into the cell-based coadds metadata. I agree that the calculation would ideally be done elsewhere, and I didn't want to make this a separate task at all, or have there be an object epoch product. However, the middleware prevents this from being done in `SourceObjectTableAnalysisTask`.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: No, this is multiple patches.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: It's in the SRD.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: This is `applyProperMotionCorrections`...can you clarify?

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: Fixed text.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: Added "to be associated" since they don't necessarily get astrometrically corrected.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: The others were matching two things that have the same provenance, and thus should have very similar positions, whereas this one is matching sources to objects. I made this a config field.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -70,11 +74,38 @@ class AssociatedSourcesTractAnalysisConnections(
         dimensions=("instrument",),
         isCalibration=True,
     )
+    properMotionCatalogs = ct.Input(
+        doc="Catalog containing proper motions.",
+        name="gbdesAstrometricFit_starCatalog",
+        storageClass="ArrowNumpyDict",
```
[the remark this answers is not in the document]
cmsaunders: I'll try that on DM-45845.

cmsaunders:
The proper motion catalog column names are going to be fixed on DM-45845. Agreed that the current format is far from ideal--it was not meant to be used by other tasks, so has never been changed from the `gbdes` default format.

For the tests requested, I would like to put those on another ticket.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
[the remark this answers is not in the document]
parejkoj: I think it's more misleading to both return and modify the input. I've been burned by it myself, getting back a value that I expected to be different from what I passed in, but they're exactly the same object. You really have to pick one or the other.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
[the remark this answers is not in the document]
parejkoj: I think this is worth pushing up to whomever is in charge of the SDM schema. I would expect that our own calculation of PM/parallax should be in our output object tables, since it is something we've said we would report. You can't fix it here, obviously, but we should be prepared to report it there eventually.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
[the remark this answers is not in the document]
parejkoj: As I think about this point, we should define somewhere (if we don't already!) what each DR's epoch will be, and then report all of the `Object` astrometric values at that epoch, then we only need the "as measured" mjd_tai timestamps for each individual measurement and the mean of that as values to feed into that correction. I don't know where the plan for this would or should go.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
parejkoj: Might want to put a TODO in this class's docstring that we expect to remove it eventually, once we are storing the measured epoch internally?

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
parejkoj: Hmmm, that's worrying to me, since the epochs won't be the same between bands. Unless we plan to shift everything to a single epoch and store all of our values relative to that? I might be confused about how this will be used.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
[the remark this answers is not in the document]
cmsaunders: There is already a long term plan for pm/parallax in the object tables, I just don't think it's going in the source tables.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: They are different between bands, which is why the `objectEpochTableTask` goes through the healsparse maps for each band, and then I'm grabbing the r-band epoch here, because I am going to compare the source positions to the r-band object positions.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
parejkoj: Ah, sorry, I meant `prepareAssociatedSources`: I was confused about which one I was looking at, because they were similar.

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -123,6 +188,77 @@ def prepareAssociatedSources(cls, skymap, tract, sourceCatalogs, associatedSourc
...
+            bandInd = dataJoined["band"] == band
+            bandSources = dataJoined[bandInd]
+            # Add key for sorting below.
+            bandSources["__index__"] = np.arange(len(bandSources))
+            bandSourcesDf = bandSources.to_pandas()
+            meanRAs = bandSourcesDf.groupby("obj_index")["coord_ra"].aggregate("mean")
+            meanDecs = bandSourcesDf.groupby("obj_index")["coord_dec"].aggregate("mean")
+
+            bandPMs = astrometricCorrectionCatalogs[band]
+            with Matcher(meanRAs, meanDecs) as m:
```
parejkoj, line 214: Having just started looking at `smatch`, I'm wondering why you're using the context manager here, instead of just calling `smatch.match`? You're only ever doing the match once with a given set of coordinates, so you're not gaining anything from creating a `Matcher` instance.

cmsaunders: I just asked Eli about this, because I've just been copying the pattern I usually see in the stack. He says there's a memory-intensive KD tree being built by `Matcher`, so it either needs to be in a context manager or explicitly deleted.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
cmsaunders: I looked at this a bit, but I don't think there's much that could be consolidated.

PR #279, opened 2024-08-09
DM-36457: Use globs instead of regexes for dataset type patterns

timj:
queryDatasetTypes no longer supports regexes.


TallJimbo (approved):
Figures that it would be code that I wrote that would be using regexes here.

PR #282, opened 2024-09-09
DM-46061: Add Metrics to calibrate task metadata

jrmullaney:


pipelines/calexpMetrics.yaml
```diff
@@ -6,3 +6,32 @@ tasks:
     config:
       atools.calexpSummaryMetrics: CalexpSummaryMetrics
       python: from lsst.analysis.tools.atools import *
+  analyzeCalibrateMetadata:
+    class: lsst.analysis.tools.tasks.MetadataAnalysisTask
+    config:
+      connections.taskName: calibrate
+      connections.outputName: calibrate_metadata
+      metadataDimensions: [instrument, visit, detector]
```
isullivan, line 14: Do these need to be `["instrument", "visit", "detector"]`?

python/lsst/analysis/tools/tasks/metadataAnalysis.py
```diff
@@ -0,0 +1,91 @@
...
+class MetadataAnalysisTask(AnalysisPipelineTask):
+    ConfigClass = MetadataAnalysisConfig
+    _DefaultName = "metadataAnalysis"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        dataId = butlerQC.quantum.dataId
+        inputs = butlerQC.get(inputRefs)
+        plotInfo = self.parsePlotInfo(inputs, dataId)
+        taskName = inputRefs.data.datasetType.name
+        taskName = taskName[: taskName.find("_")]
```
isullivan, line 86: Not terribly important, but you might as well move the `taskName` definition inside the `if not metadata:` block below.

isullivan (approved):
Looks good. We should try to replace both `metadataExposureDetectorAnalysis.py` and `diffimTaskDetectorVisitMetricsAnalysis.py` with calls to your new generic class. Those can be done on new tickets, though.

python/lsst/analysis/tools/tasks/metadataAnalysis.py
```diff
@@ -0,0 +1,91 @@
...
+class MetadataAnalysisConnections(
+    AnalysisBaseConnections,
+    dimensions={"instrument", "exposure", "detector"},
+    defaultTemplates={"taskName": "isr"},
+):
+    data = connectionTypes.Input(
+        doc="Task metadata to load.",
+        name="{taskName}_metadata",
+        storageClass="TaskMetadata",
+        deferLoad=True,
```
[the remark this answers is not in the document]
jrmullaney: `deferLoad` is needed for the `plotInfo = self.parsePlotInfo(inputs, dataId)` in `def runQuantum(...` below. Without it, `tableName = inputs[connectionName].ref.datasetType.name` in `parsePlotInfo` returns the attribute error `'TaskMetadata' object has no attribute 'ref'`.

PR #283, opened 2024-09-12
DM-45843: Create a first attempt at a metric information yaml file

sr525:


mrawls (approved):
Please add the missing newline which is entirely my fault

PR #289, opened 2024-09-27
DM-45899: Export TreecorrConfig to meas_algorithms for TEx computation

PFLeget:
In order to compute TEx, I need `TreecorrConfig` to be in `meas_algorithms`, so I export it to `meas_algorithms`.

### DIFF
--- a/python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
+++ b/python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
@@ -21,17 +21,15 @@
 
 from __future__ import annotations
 
-__all__ = (
-    "CalcRhoStatistics",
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -21,17 +21,15 @@
 
 from __future__ import annotations
 
-__all__ = (
-    "CalcRhoStatistics",
-    "TreecorrConfig",
```
arunkannawadi, line 26: Because we committed to `TreecorrConfig` being available from here, trying to import `TreecorrConfig` from here should still continue to work (with a deprecation warning) until the next major release.

PR #291, opened 2024-10-02
DM-44545: Design initial calibration report format

czwa:
Fix missing PTC dataset type.

### DIFF
--- a/python/lsst/analysis/tools/actions/plot/elements/scatterElement.py
+++ b/python/lsst/analysis/tools/actions/plot/elements/scatterElement.py
@@ -104,7 +104,7 @@ class ScatterElement(PlotElement):
             data[self.xKey] if self.xKey is not None else range(len(data[self.valsKey])),  # type: ignore
             data[self.valsKey],  # type: ignore
             color=self.color if self.color is not None else None,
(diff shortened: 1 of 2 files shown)


sr525 (approved):

python/lsst/analysis/tools/actions/plot/elements/scatterElement.py
```diff
@@ -104,7 +104,7 @@ def __call__(self, data: KeyedData, ax: Axes, **kwargs) -> KeyedData:
             data[self.xKey] if self.xKey is not None else range(len(data[self.valsKey])),  # type: ignore
             data[self.valsKey],  # type: ignore
             color=self.color if self.color is not None else None,
-            marker=self.marker if self.marker is not None else None,
+            marker=self.marker if self.marker is not None else ".",
```
sr525, line 107: I agree that a lot of these defaults seem a little weird.

czwa: There was some pushback to this change on slack, but I plan to still merge it with this logic:  merging this change will make some line plots that use this element look different;  not merging it will make scatter plots that use this element empty.

PR #294, opened 2024-10-15
DM-46684: Add Metrics to calibrateImage task metadata

jrmullaney:


pipelines/visitQualityCore.yaml
```diff
@@ -46,3 +46,87 @@ tasks:
     config:
       atools.astromColorDiffMetrics: TargetRefCatDeltaColorMetrics
       python: from lsst.analysis.tools.atools import *
+  calexpSummary:
+    class: lsst.analysis.tools.tasks.CalexpSummaryAnalysisTask
+    config:
+      atools.calexpSummaryMetrics: CalexpSummaryMetrics
```
isullivan, line 52: Once DM-43077 merges, there won't be any more references to a `calexp`. We should rethink these names, perhaps use `pviSummary` and `pviSummaryMetrics`. Maybe get some naming suggestions from #dm-naming-things, since I don't love the `pvi` suggestions either.

python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -41,9 +41,11 @@ class MetadataMetricTool(AnalysisTool):
...
     )
 
-    subTaskName = Field[str](
-        doc="The name of the subtask to extract metadata from. "
-        "If None, the entire task metadata will be used.",
+    subTaskNames = DictField[str, str](
+        doc="The names of the subtasks to extract metadata from. "
+        "If the metric name is identified as one of the keys, then "
+        "the corresponding value is taken as the subTask metadata "
+        "from which to extract metadata.",
```
isullivan, line 48: The old comment is I think still also helpful: `"If None, the entire task metadata will be used."`

pipelines/calexpMetrics.yaml
```diff
@@ -1,37 +0,0 @@
-description: |
```
isullivan, line 1: Can this be deprecated instead of removed?

isullivan (approved):
Looks good, but please take the `Calexp` metric names to #dm-naming-things to get broader feedback on those names.

python/lsst/analysis/tools/atools/metadataMetrics.py
jrmullaney: I think now that the old comment was incorrect. I now think that whether it uses the whole metedata depends on whether the metadata is saved as JSON or YAML. Not ideal, since one often doesn't know how metadata is saved since we let the butler take care of that. I've not looked into it deeply, though, so that's just my hunch atm.

Either way, I know that in some instances it doesn't search the whole metadata if subTaskNames is None, so this system is safer.

pipelines/calexpMetrics.yaml
jrmullaney: Yes - I've reinstated calexpMetrics.yaml, added a note to its description, and have created a ticket to remove it eight weeks from now (as per the DM guidelines).

pipelines/visitQualityCore.yaml
jrmullaney: I've left the names as the are for now, but have created a ticket (DM-46932) to rename/delete/whatever as part of DM-43077.

PR #298, opened 2024-10-21
DM-46918: Updates to use the_monster as a refcat

psferguson:


python/lsst/analysis/tools/tasks/photometricCatalogMatch.py
```diff
@@ -84,6 +84,8 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         # the HSC ones do, this is a bit hacky but fixes this
         if bands[0].startswith("lsst") or "sim" in bands[0] or "smeared" in bands[0]:
             bands = self.config.filterNames
+        elif bands[0].startswith("monster"):
+            bands = [band[-1] for band in bands]
```
erykoff, line 88: Why the -1?  I don't fully follow this code (or why everything is so hacked up...)

erykoff (commented):
Can you add a bit more about what these fields/variables mean?  (I know this is mostly pre-existing but it's definitely confusing).

python/lsst/analysis/tools/tasks/photometricCatalogMatch.py
psferguson: I added some more documentation, is this enough?

erykoff: Ah that helps.  This is ticketed to fix this correctly, right?  And fortunately we only have one character bands in the monster (maybe worth mentioning that we don't support NB in the monster).

erykoff (approved):

python/lsst/analysis/tools/tasks/photometricCatalogMatch.py
psferguson: Added the ticket number to fix this

PR #300, opened 2024-10-22
DM-47046: Return _ref to matched difference metric name suffix

taranu:


python/lsst/analysis/tools/atools/diffMatched.py
```diff
@@ -663,6 +663,7 @@ def setDefaults(self):
 
         self.mag_bins_plot.mag_interval = 100
         self.mag_bins_plot.mag_width = 200
+        self.name_suffix = "_mag{name_mag}"
```
jrmullaney, line 666: Maybe leave a comment (referencing the docstring?) to explain why this is needed.

jrmullaney (approved):

PR #301, opened 2024-10-24
DM-47111: Fix units in analysis_tools/.../calexpMetrics.py

jrmullaney:
Where possible, units now align with those in afw/image/_exposureSummaryStats.py

### DIFF
--- a/python/lsst/analysis/tools/atools/calexpMetrics.py
+++ b/python/lsst/analysis/tools/atools/calexpMetrics.py
@@ -35,39 +35,41 @@ class CalexpSummaryMetrics(AnalysisTool):
 
     # raCorners and decCorners statistics cannot be written to a metric,
     # as metrics can only be single-valued (i.e., scalars).
+    # Units in comments are to indicate compound units, which are currently
+    # unsupported.
     _units = {
         "psfSigma": "pixel",
-        "psfArea": "",  # There is no astropy sq. pixel unit.
-        "psfIxx": "pixel",
-        "psfIyy": "pixel",
-        "psfIxy": "pixel",
+        "psfArea": "",  # pixel**2
+        "psfIxx": "",  # pixel**2
+        "psfIyy": "",  # pixel**2
+        "psfIxy": "",  # pixel**2
         "ra": "degree",
         "dec": "degree",
-        "pixelScale": "",  # There is no astropy arcsec/pixel unit.
+        "pixelScale": "",  # arcsec/pixel.
         "zenithDistance": "degree",
         "expTime": "s",
         "zeroPoint": "mag",
-        "skyBg": "mag",
-        "skyNoise": "mag",
-        "meanVar": "mag",
+        "skyBg": "electron",
+        "skyNoise": "electron",
+        "meanVar": "",  # electron**2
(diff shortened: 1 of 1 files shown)


kadrlica (changes requested):
Only one comment about units for flux units (related to sky and variance). They were in ADU up until recently, where they got switched to electron by new ISR (i.e., the great calibration push for ComCam). I think that we should probably "code forward" and use electron as the unit, with the understanding that this will be more useful for ComCam and beyond.

python/lsst/analysis/tools/atools/calexpMetrics.py
```diff
@@ -35,39 +35,41 @@ class CalexpSummaryMetrics(AnalysisTool):
...
+        "pixelScale": "",  # arcsec/pixel.
         "zenithDistance": "degree",
         "expTime": "s",
         "zeroPoint": "mag",
-        "skyBg": "mag",
-        "skyNoise": "mag",
-        "meanVar": "mag",
+        "skyBg": "adu",
+        "skyNoise": "adu",
+        "meanVar": "",  # adu**2
```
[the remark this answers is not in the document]
jrmullaney: Thanks, @kadrlica, makes sense. These have now been changed to `electron`.

python/lsst/analysis/tools/atools/calexpMetrics.py
```diff
@@ -35,39 +35,41 @@ class CalexpSummaryMetrics(AnalysisTool):
...
+        "skyNoise": "electron",
+        "meanVar": "",  # electron**2
         "astromOffsetMean": "arcsec",
         "astromOffsetStd": "arcsec",
-        "nPsfStar": "",
-        "psfStarDeltaE1Median": "pixel",
-        "psfStarDeltaE2Median": "pixel",
-        "psfStarDeltaE1Scatter": "pixel",
-        "psfStarDeltaE2Scatter": "pixel",
+        "nPsfStar": "ct",
```
kadrlica, line 57: While I originally thought this usage of "count" was correct, on reading more documentation it seems like it is used more as a photometric unit, so it might make more sense to leave this unitless. Here is some usage I found [in the astropy doc](https://docs.astropy.org/en/latest/units/ref_api.html#module-astropy.units.function.units) and some [remarks in the iova doc](https://www.ivoa.net/documents/VOUnits/20231215/REC-VOUnits-1.1.html#sec:other).  I'd be interested in your opinion.

kadrlica (approved):
This looks good to me. I left one question about the meaning of the unit "count", but I'm happy to have this merged regardless of what you decide on that.

python/lsst/analysis/tools/atools/calexpMetrics.py
jrmullaney: Yeah - this has crossed my mind in the past. I've used it in other places to refer to things that are unitless integer and "counting". GIven that I (and others) have used it in other places with this meaning, I'd like to keep it here. However, I think it would be useful to do an audit of atools to make sure that it is used consistently.

kadrlica: Sounds good to me!

PR #302, opened 2024-10-24
DM-46188: Add connection templates to refCatSourcePhotometricAnalysis.

TallJimbo:
All other refCat*Analysis tasks have these, so this one should do.

### DIFF
--- a/python/lsst/analysis/tools/tasks/refCatSourcePhotometricAnalysis.py
+++ b/python/lsst/analysis/tools/tasks/refCatSourcePhotometricAnalysis.py
@@ -30,11 +30,15 @@ from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPi
 class RefCatSourcePhotometricAnalysisConnections(
     AnalysisBaseConnections,
     dimensions=("visit",),
-    defaultTemplates={"outputName": "sourceTable_visit_ps1_pv3_3pi_20170110_match_photom"},
+    defaultTemplates={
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/tasks/refCatSourcePhotometricAnalysis.py
```diff
@@ -30,11 +30,15 @@
 class RefCatSourcePhotometricAnalysisConnections(
     AnalysisBaseConnections,
     dimensions=("visit",),
-    defaultTemplates={"outputName": "sourceTable_visit_ps1_pv3_3pi_20170110_match_photom"},
+    defaultTemplates={
+        "targetCatalog": "sourceTable_visit",
+        "refCatalog": "ps1_pv3_3pi_20170110",
+        "outputName": "sourceTable_visit_ps1_pv3_3pi_20170110_match_photom",
```
erykoff, line 36: Oh is there a way that the output name (where is this used, in the parent?) can use the templates?

erykoff (approved):
Small question, maybe the answer is "no"

python/lsst/analysis/tools/tasks/refCatSourcePhotometricAnalysis.py
TallJimbo: I _think_ the answer is no, but I wish it were otherwise.  This is my first encounter with a lot of this stuff so I'm not certain.

PR #306, opened 2024-11-06
DM-47010: Change default to assume preSources used for matching

TallJimbo:
Use isolated_star_presources for matches and sourceTable_visit for measurements.


arunkannawadi (approved):
LGTM! It could use an update to the comment here to reflect this change:
https://github.com/lsst/analysis_tools/blob/b70190bcbfeb07443d799251cc509ea0780443d5/python/lsst/analysis/tools/atools/astrometricRepeatability.py#L222

PR #308, opened 2024-11-14
DM-42201: Add flat metrics for mean, noise vs mjd and FP plots

Alex-Broughton:


python/lsst/analysis/tools/atools/calibQuantityProfile.py
```diff
@@ -90,16 +91,19 @@ class SingleValueRepacker(KeyedDataAction):
 
     def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
         repackedData = {}
-        uniquePanelKeys = list(set(data[self.panelKey]))
+        good = ~data[self.panelKey].mask
```
czwa, line 94: Unfortunately, I think (based on your jenkins failure), that you need to see if `data` is a numpy masked-array, and do the new thing, otherwise, default to the existing code.  Hopefully that's not too difficult, and I'll re-check next week.

czwa (changes requested):

pipelines/cpCore.yaml
```diff
@@ -139,12 +139,31 @@ tasks:
...
+      atools.flatNoisePerAmp: CalibStatisticFocalPlanePlot
+      atools.flatNoisePerAmp.produce.plot.addHistogram: true
+      atools.flatNoisePerAmp.quantityKey: FLAT_NOISE
+      atools.flatNoisePerAmp.unit: electron
+
+      python: |
+        from lsst.analysis.tools.atools import *
+
+  analyzeFlatDetMergeCore:
+    class: lsst.analysis.tools.tasks.VerifyCalibAnalysisTaskByFilter
```
czwa, line 162: This is calling the existing task that only has dimensions `["instrument", "physical_filter"]`.  We need a new task that has input data dimensions of `["instrument", "physical_filter", "detector"]`.

czwa: I think this means we need to add one more task with that set of dimensions.  This all needs to be cleaned up next year to avoid this problem.

czwa (changes requested):

czwa (approved):
Approved, provided Jenkins is successful, as the "changes requested" appear to be already implemented.

czwa:
Same comment as on the other PR:  the rebase should get rid of the `dist-info` file via `gitignore`, otherwise it should be dropped manually.  Otherwise, fine.

PR #309, opened 2024-11-15
DM-47544: Change calcSizeResidual from (T_obs-T_psf)/T_psf to (T_obs-T_psf)/T_obs

cahebert:


PFLeget (approved):
Looks the same as what is done in Piff now!

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -248,7 +248,7 @@ def __call__(self, data: KeyedData, **kwargs) -> KeyedData:
         eRes = calcEDiff(data, **kwargs)
         eRes /= responsitivity  # type: ignore
         e1Res, e2Res = np.real(eRes), np.imag(eRes)
-        sizeRes = calcSizeResidual(data, **kwargs)
+        sizeRes = -1 * calcSizeResidual(data, **kwargs)  # sign flip residual to T_psf - T_model
```
arunkannawadi, line 251:
```suggestion
        sizeRes = calcSizeResidual(data, **kwargs)
        sizeRes *= -1  # sign flip residual to T_psf - T_model
```

arunkannawadi (commented):
I think the sizeRes calculation need to happen in two lines as suggested, and that should make the Jenkins pass. I also wonder if dividing by the measured size requires some addition check about it being zero, but that's likely taken care of upstream. Let us know if you need help with any of the workflows @cahebert . It'd be great if we could merge this by this Wednesday.

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
cahebert: So sorry, this completely fell off my radar! I double checked and yes, the division is taken care of upstream.

The last Jenkins run I did of this ([here](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/3245/pipeline)) did pass — does your suggested change still stand?

arunkannawadi:
Merging this without any further delay. Jenkins is green on all three OSs.

PR #310, opened 2024-11-18
[DM-46910] Add unit test for calcRhoStatistics.

PFLeget:


tests/test_actions.py
```diff
@@ -138,6 +139,45 @@ def setUp(self):
...
+        sizeRho = 1000
+        size_src = np.random.normal(scale=1e-3, size=sizeRho)
+        e1_src = np.random.normal(scale=1e-3, size=sizeRho)
+        e2_src = np.random.normal(scale=1e-3, size=sizeRho)
+
+        size_psf = np.random.normal(scale=1e-3, size=sizeRho)
+        e1_psf = np.random.normal(scale=1e-3, size=sizeRho)
+        e2_psf = np.random.normal(scale=1e-3, size=sizeRho)
+
+        def _getMatrixElement(size, e1, e2):
```
arunkannawadi, line 152: Could you put this outside of `setUp` may be as a `staticmethod`?

arunkannawadi, line 152:
```suggestion
        def _getMatrixElements(size, e1, e2):
```

tests/test_actions.py
```diff
@@ -174,7 +214,20 @@ def testCalcBinnedStats(self):
...
+        rho.treecorr.min_sep = 0.01
+        rho.treecorr.max_sep = 100.0
+        rho.treecorr.sep_units = "arcmin"
+        result = rho(self.dataRhoStats, band="r")
+        for rho in result:
+            if rho != "rho3alt":
+                assert np.sum(np.isfinite(result[rho].xip)) == len(result[rho].xip)
+                assert np.sum(np.isfinite(result[rho].xim)) == len(result[rho].xim)
+            else:
+                assert np.sum(np.isfinite(result[rho].xi)) == len(result[rho].xi)
```
arunkannawadi, line 230:
```suggestion
                self.assertEqual(np.isfinite(result[rho].xi)), len(result[rho].xi))
```

and similar for the ones above, so you get a more helpful error message if/when the test fails.

tests/test_actions.py
```diff
@@ -138,6 +139,45 @@ def setUp(self):
 
         self.data = pd.DataFrame.from_dict(data)
 
+        np.random.seed(42)
```
arunkannawadi, line 142: I presume for this particular seed none of the ellipticity magnitudes exceed 1? If so, it might be good to state that in a comment, saying that there's no guard against such a thing.

tests/test_actions.py
arunkannawadi: Or, `self.assertEqual(len(result[rho].xi), rho.treecorr.nbins)` and `self.assertTrue(np.isfinite(result[rho].xi).all())`

arunkannawadi (approved):
Thanks for adding this much-needed test. One general suggestion would be to just create a new class since that `setUp` required for rhoStatistics seem to have no overlap whatsoever with the `setUp` required for other tests here.

PR #313, opened 2024-11-21
DM-47629: Generate metric aggregation plots in analysis_tools

jrmullaney:


python/lsst/analysis/tools/actions/plot/histPlot.py
```diff
@@ -608,12 +608,16 @@ def _addStatisticsPanel(
         legend_labels = (
             ([""] * (len(handles) + 1))
             + [stats_dict["statLabels"][0]]
-            + [f"{x:.3g}" for x in stats_dict["stat1"]]
+            + [f"{x:.3g}" if x > 0.01 else f"{x:.2e}" for x in stats_dict["stat1"]]
```
arunkannawadi, line 611: Should this condition on the absolute value of `x` instead?

python/lsst/analysis/tools/actions/plot/histPlot.py
```diff
@@ -608,12 +608,16 @@ def _addStatisticsPanel(
...
             + [stats_dict["statLabels"][1]]
-            + [f"{x:.3g}" for x in stats_dict["stat2"]]
+            + [f"{x:.3g}" if x > 0.01 else f"{x:.2e}" for x in stats_dict["stat2"]]
             + [stats_dict["statLabels"][2]]
-            + [f"{x:.3g}" for x in stats_dict["stat3"]]
+            + [f"{x:.3g}" if x > 0.01 else f"{x:.2e}" for x in stats_dict["stat3"]]
         )
+        # Replace "e+0" with "e" and "e-0" with "e-" to save space.
+        legend_labels = [label.replace("e+0", "e") for label in legend_labels]
```
arunkannawadi, line 618: Is there a reason to include the 0 after the `+` sign here? I'd have imagined just replacing `e+` by `e`. Also if the value is say `2.32432432423`, it gets formatted first as `2.32e+00` above and it gets modified to `2.32e0` which is still not as clean as it could be. I wonder if a regex substitution might be more robust here.

python/lsst/analysis/tools/atools/calexpMetrics.py
```diff
@@ -77,3 +84,24 @@ def setDefaults(self):
...
+class CalexpMetricHists(AnalysisTool):
+    """
+    Class to generate histograms of metrics extracted from a Metrics Table.
+    One plot per band.
+    """
+
+    parameterizedBand: bool = False
+    metrics = DictField[str, str](doc="The metrics to plot and their respective labels.")
+
+    def setDefaults(self):
```
arunkannawadi, line 98:
```suggestion
    def setDefaults(self):
        super().setDefaults()
```

python/lsst/analysis/tools/tasks/metricAnalysis.py
```diff
@@ -26,33 +26,88 @@
...
 
     ConfigClass = MetricAnalysisConfig
     _DefaultName = "metricAnalysis"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        """Take a metric table and run an analysis tool on the
+        data it contains. This could include creating a plot
+        the metrics and/or calculating summary values of those
+        metrics, such as means, medians, etc. The analysis
+        is outlined within the analysis tool.
```
arunkannawadi, line 89: I would just putting this in doc for the class, and replacing this entire docstring with a comment `# Docstring inherited.`

arunkannawadi (approved):

python/lsst/analysis/tools/actions/plot/histPlot.py
jrmullaney: Yes, it's to shorten it as much as possible. The 0 in e+0 is redundant, so I remove it.
Wrt the 2.324... conversion. I'm not getting what you say. I'm just getting 2.32 at the end. The g format code converts it to 2.32, not 2.32e+00.

arunkannawadi: OK, I probably skipped an earlier formatting when I was checking out this specific line then.

PR #314, opened 2024-11-22
DM-47633: Remove bands appearing in plots' names

arunkannawadi:


jrmullaney (approved):
This all looks fine to me, and I've managed to get it to work for the metric plots I was making last week, so we can supersede that ticket with this one.

The only slight issue I encountered was with the `plotInfo` displaying all six bands on the plot info panel when, in reality, the data shown is for a single band. I can fix this by passing `bands` as a kwarg in my task's call to `self.run`, but I was wondering whether it would be better to default to setting the plotInfo to include only the current dimension band. Does that make sense - I probably haven't explained it very well.

arunkannawadi:
It makes sense to me, but should that necessarily be part of this ticket?

I'm seeing single-visit plots have all the bands listed [here](https://usdf-rsp-dev.slac.stanford.edu/plot-navigator/plot/embargo/LSSTComCam%2Fruns%2FnightlyValidation%2F20241122%2Fw_2024_46%2FDM-47059/preSourceTableCore_skySourceSky_SkyPlot) even without this ticket. Did you notice any such thing appearing in the plots as a result of the changes in this ticket?

If I were to do this, I'd have read the `band` in `parsePlotInfo` defined [here](https://github.com/lsst/analysis_tools/blob/edc5d771af79b4394a306b08e8198a212c8e813c/python/lsst/analysis/tools/interfaces/_task.py#L470-L501) but that is overridden [here](https://github.com/lsst/analysis_tools/blob/edc5d771af79b4394a306b08e8198a212c8e813c/python/lsst/analysis/tools/interfaces/_task.py#L408). Given that various tasks implement their own `run` and `parsePlotInfo` method, I'd like to defer this to another ticket so there'd be an opportunity to prevent this more comprehensively, but agree with your desire to not have to pass `bands` as a `kwarg`.

PR #315, opened 2024-11-25
DM-41181: Add survey-wide healsparse property map plots to analysis_tools

enourbakhsh:


python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -41,6 +45,116 @@
...
+        [mpl_path_effects.Stroke(linewidth=4, foreground="white", alpha=0.8), mpl_path_effects.Normal()]
+    )
+
+
+def prettyPrintFloat(n):
+    if n.is_integer():
+        return str(int(n))
+    return str(n)
+
+
```
enourbakhsh, line 157: Here, I simply moved pre-existing static methods outside the class to make them easier to reuse for other tasks in the future.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -339,14 +453,12 @@ def makePlot(
                         rcparams=rcparams,
                     )
                     sp.draw_hspmap(mapData, zoom=zoom)
-                    sp.set_xlabel("RA")
-                    sp.set_ylabel("Dec")
+                    sp.ax.set_xlabel("RA")
```
enourbakhsh, line 456: Quick fix for a SkyProj warning due to a recent update.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -41,6 +45,110 @@
...
+        Zoom factor; for instance, n=2 means zooming in 2 times at the
+        center. If None, the function returns None.
+
+    Returns
+    -------
+    `tuple` [`float`]
+        New extent as (new_lon_min, new_lon_max, new_lat_min, new_lat_max).
+    """
+    if n is None:
+        return None
```
sr525, line 75: Why is this necessary?

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -506,112 +614,232 @@ def getOutputNames(self, config=None) -> Iterable[str]:
...
+            "xtick.labelsize": 13,
+            "ytick.labelsize": 13,
+        }
+
+        mapName = data["data"].ref.datasetType.name
+        mapData = data["data"].get()
+
+        with plt.rc_context(rcparams):
+            # The figsize should be decided based on the survey footprint.
+            fig, ax = plt.subplots(1, 1, figsize=(19, 7))
```
sr525, line 755: Can you try removing the plt? The code is pretty slow and it would be useful to speed it up. The property map code in general is slow.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
enourbakhsh: We use this function within a loop where we occasionally pass `None` to indicate that the full original plot should be displayed. Skyproj interprets `extent=None` as a signal that no zooming is required and renders the complete plot.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
enourbakhsh: Is the `with plt.rc_context(rcparams)` part causing the slow-down? If I remember right, this is how @erykoff  suggested setting `rcparams`. I mentioned him to see if he has any thoughts.

erykoff: What @sr525 is saying is that `plt.subplots` is *bad*. You want to instead use something like this https://github.com/lsst/analysis_tools/blob/eed07097883c01ff4d95e4b172ae7cabdcadfd7d/python/lsst/analysis/tools/actions/plot/skyPlot.py#L219-L220

erykoff: Regarding the `plt.rc_context()` that is fine, but you should instead do `from matplotlib import rc_context`.  You should not ever do `import matplotlib.pyplot as plt` in any pipeline code if at all possible.

enourbakhsh: Ah, got it! I was confused because `rcparams` came up on Jira.

enourbakhsh: I made those changes to see if it helps speed things up. The `make_fig` approach caused an error downstream:
 
 ```
  File "/home/en7908/repos/analysis_tools/python/lsst/analysis/tools/actions/plot/propertyMapPlot.py", line 827, in makePlot
    cbar = sp.draw_colorbar(**{"location": "top", "pad": 0.2, **plotConfig.colorbarKwargs})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/HSC/LSST/stack_20240827/conda/envs/lsst-scipipe-9.0.0/lib/python3.11/site-packages/skyproj/_skyproj.py", line 1486, in draw_colorbar
    cbar = plt.colorbar(ax=ax, location=location, ticks=ticks,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 RuntimeError: No mappable was found to use for colorbar creation. First define a mappable such as an image (with imshow) or a contour set (with contourf).
 ```
 Looks like it doesn't like the axis created by `ax = fig.add_subplot(111)`.

erykoff: This should never have worked, it was just a coincidence.  This should be `ax=sp.ax` to get the correct axis.

erykoff: Also you do not want to use `plt.colorbar()`.  Hold on ...

erykoff: You can use `fig.colorbar()`.  Don't use `plt` at all!  Anywhere you have that it is going to be a problem.

enourbakhsh: The `plt.colorbar()` line is from `_skyproj.py`.

erykoff: Ha!  Okay, well I have that fixed on skyproj 2 (beta), which I'll be testing with the stack after commissioning is over.  Sorry about that!  But this error is now coming from within skyproj?  Ugh, hold on...

sr525: Thank you for the help Eli.

PR #320, opened 2024-12-02
DM-47892 : Add associationCore

Gerenjie:


pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -79,6 +79,19 @@ tasks:
...
+    class: lsst.analysis.tools.tasks.MetadataAnalysisTask
+    config:
+      connections.taskName: diaPipe
+      connections.outputName: diaPipe_metadata
+      metadataDimensions: ["instrument", "visit", "detector"]
+      atools.calexpMetadataMetrics: MetadataMetricTool
+      atools.calexpMetadataMetrics.taskName: diaPipe
+      atools.calexpMetadataMetrics.metrics:
+        nExpectedSsObjects: ct
+        nAssociatedSsObjects: ct
```
isullivan, line 92: It looks like these quantities were added to the task metadata previously in a different ticket, so we should use those names:
```
numTotalSolarSystemObjects
numAssociatedSsObjects
```

pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -79,6 +79,19 @@ tasks:
...
         from lsst.analysis.tools.atools import DiffimMetadataMetricTool
+  associationCore:
+    class: lsst.analysis.tools.tasks.MetadataAnalysisTask
+    config:
+      connections.taskName: diaPipe
+      connections.outputName: diaPipe_metadata
+      metadataDimensions: ["instrument", "visit", "detector"]
+      atools.calexpMetadataMetrics: MetadataMetricTool
+      atools.calexpMetadataMetrics.taskName: diaPipe
+      atools.calexpMetadataMetrics.metrics:
```
isullivan, line 90: `atools.calexpMetadataMetrics` as used here is just a label, which references `MetadataMetricTool` and allows it to be configured. You can name it anything you want, so I suggest renaming to `atools.diaPipeMetadataMetrics`

isullivan (approved):
Some naming changes requested.

PR #324, opened 2024-12-06
DM-47989: Add unit test for survey-wide healsparse property map task/tool/action

enourbakhsh:


pipelines/surveyQualityCore.yaml
```diff
@@ -5,12 +5,14 @@ tasks:
     class: lsst.analysis.tools.tasks.PropertyMapSurveyWideAnalysisTask
     config:
       connections.outputName: propertyMapSurveyWide
+      autozoom: true
       projection: McBryde
       projectionKwargs:
+        celestial: true
```
jrmullaney, line 11: What's the need for this, given that the ticket is to just make a test suite?
(I'm just checking that something hasn't sneaked-in by mistake)

jrmullaney (approved):
I've just added something for you to double check, in case it's sneaked-in to the commit by mistake.

pipelines/surveyQualityCore.yaml
enourbakhsh: I made that change along with a few minor editorial updates since opening another ticket didn’t feel necessary. Just to clarify, those are default configurations, and I wanted to make sure users know they can modify those parameters.

PR #325, opened 2024-12-07
DM-48019: Add explicit input for sourceTable_visit in analyzeSourceTableCore

enourbakhsh:
This avoids mistakenly leaving the default input unchanged when reusing for `preSources` (instead of `sources`).

### DIFF
--- a/pipelines/visitQualityCore.yaml
+++ b/pipelines/visitQualityCore.yaml
@@ -4,6 +4,7 @@ tasks:
   analyzeSourceTableCore:
     class: lsst.analysis.tools.tasks.SourceTableVisitAnalysisTask
     config:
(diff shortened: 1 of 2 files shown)


pipelines/visitQualityCore.yaml
```diff
@@ -4,6 +4,7 @@ tasks:
   analyzeSourceTableCore:
     class: lsst.analysis.tools.tasks.SourceTableVisitAnalysisTask
     config:
+      connections.inputName: sourceTable_visit
```
yalsayyad, line 7: Technically not necessary if the task default has `defaultTemplates={"inputName": "sourceTable_visit"}`

yalsayyad (approved):

PR #326, opened 2024-12-08
DM-48024: Modify metadata getter for metadata metrics

enourbakhsh:


python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -64,8 +64,10 @@ def validate(self):
 class DatasetMetadataMetricTool(MetadataMetricTool):
     """Tool designed to extract values from metadata of data products"""
 
-    metricsStoredAsDict = DictField[str, bool](
```
leeskelvin, line 67: As this is a changed config, does this need to be deprecated?

python/lsst/analysis/tools/atools/metadataMetrics.py
```diff
@@ -64,8 +64,10 @@ def validate(self):
...
-        doc="Whether metrics are stored as a dictionary using `set_dict()`.",
+    metricsPrefixedWithBaseKeys = DictField[str, bool](
+        doc="Whether metrics are prefixed with base keys. For each key (a metric name or base key), "
+        "the corresponding boolean value specifies if the metric should be extracted using the base "
+        "key as a prefix.",
         default=None,
```
leeskelvin, line 71: Why does this default to `None` rather than `False`?

leeskelvin:
Also mostly good here. Questions above about deprecation and defaults. As with the `ip_isr` ticket, please also squash this PR down into a single commit.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: I just introduced this config last week for a specific ticket. I did a quick search, and no one else is using it (even if they were, it wouldn’t have worked for the reasons discussed in this ticket). So, I think it’s fine to remove it without deprecation.

python/lsst/analysis/tools/atools/metadataMetrics.py
enourbakhsh: I saw examples like [this one](https://github.com/lsst/analysis_tools/blob/e452f7d79f336b7d8be42b4ac7a24ac7647eb724/python/lsst/analysis/tools/actions/vector/vectorActions.py#L285) and followed them. However, I also found examples with `default={}`, which makes more sense to me as well. I’ve updated it to `{}`.

enourbakhsh: I had multiple failed Jenkins runs (one error occurred in t`ests/test_cliCmdRetrieveArtifacts.py`; more details on Jira) after making this change. I switched it back to `default={}`, but got the same Jenkins error. So, this wasn’t the cause. Essentially, I got a failed Jenkins run on the very same PRs that previously passed Jenkins successfully (except that I squashed the commits into a single commit). I'm confused!

enourbakhsh: Waited for a bit and ran another Jenkins to see if it was an intermittent failure. Updates on Jira.

leeskelvin: Can you test again with `dafault={}`, if this was intermittent?

enourbakhsh: Two Jenkins failed with `default={}` again. Back to `default=None`.

leeskelvin: Ok, I don't want to hold this up any longer, so please go ahead and merge.

enourbakhsh: My precautionary Jenkins runs with `default=None` also failed two times in a row. Very strange.

leeskelvin: Rather than waiting for Jenkins, can you reproduce running the failing unit test locally?

enourbakhsh: Can't reproduce it locally.

PR #330, opened 2024-12-12
DM-34875: handle input tables written originally as ArrowAstropy rather than DataFrame

TallJimbo:


python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
```diff
@@ -435,10 +436,11 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         for refCatRef in inputs["refCat"]:
             refCat = refCatRef.get(
                 parameters={
-                    "columns": ["detect_isPrimary", self.config.ra_column, self.config.dec_column]
+                    "columns": ["detect_isPrimary", self.config.ra_column, self.config.dec_column, "objectId"]
                     + refCatSelectors
                 }
             )
+            refCat.set_index("objectId")
```
erykoff, line 443: Is this setting an astropy table index?

erykoff (approved):
A couple of small questions

python/lsst/analysis/tools/tasks/associatedSourcesTractAnalysis.py
```diff
@@ -164,7 +164,7 @@ def prepareAssociatedSources(
         """Concatenate source catalogs and join on associated object index."""
 
         # Keep only sources with associations
-        sourceCatalogStack = vstack(sourceCatalogs)
+        sourceCatalogStack = vstack(sourceCatalogs, join_type="exact")
```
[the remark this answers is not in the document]
TallJimbo: The join types all work the same if all of the input catalogs have the same columns.  Switching to `exact` just makes it an error if they don't, instead of adding new columns with NaNs.

python/lsst/analysis/tools/tasks/sourceObjectTableAnalysis.py
TallJimbo: No this one's still a `DataFrame`, because there was a lot of `pandas`-heavy code using it.  I'm setting the index here because the thing that's still being read in as a `DataFrame` is now being written as `ArrowAstropy` and hence isn't getting the index set on write, as the code here previously expected.

PR #331, opened 2024-12-13
DM-48137: Remove units from metric when writing to metric table

jrmullaney:


python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -184,9 +186,10 @@ def run(self, dataIdInfo, metricBundles, skymap):
         # Add the metrics from the first bundle to the dict
         for name, metrics in metricBundles[0].items():
             for metric in metrics:
-                fullName = f"{name}_{metric.metric_name}"
-                metricValue = metric.quantity
-                metricsDict[fullName] = [metricValue]
+                fullName = f"{metric.metric_name}"
+                metricsDict[fullName] = [metric.quantity.value]
+                if metric.quantity.unit is not u.dimensionless_unscaled and metric.quantity.unit is not u.pct:
```
jrmullaney, line 191: Astropy Tables don't like either `dimensionless` or `%` units.

python/lsst/analysis/tools/tasks/makeMetricTable.py
```diff
@@ -181,12 +184,14 @@ def run(self, dataIdInfo, metricBundles, skymap):
...
+        # Add the metrics and units from the first bundle to the dicts
         for name, metrics in metricBundles[0].items():
             for metric in metrics:
                 fullName = f"{name}_{metric.metric_name}"
-                metricValue = metric.quantity
-                metricsDict[fullName] = [metricValue]
+                metricsDict[fullName] = [metric.quantity.value]
+                # "Dimensionless" and percent units not allowed in Tables:
+                if metric.quantity.unit is not u.dimensionless_unscaled and metric.quantity.unit is not u.pct:
+                    metricUnits[fullName] = metric.quantity.unit
```
kherner, line 194: Would it make sense to raise some error here if the unit were not of an allowed type? At the moment, it's just as no-op.

kherner (approved):

python/lsst/analysis/tools/tasks/makeMetricTable.py
jrmullaney: Thanks for querying this. `dimensionless_unscaled` just means the metric doesn't have a unit associated with it, so there's nothing to put in the table anyway. In that case an error or warning isn't appropriate - it would effectively be raising an error to say that a metric doesn't have a unit
In the case of percent, I think a warning would be more appropriate; I don't think we want to stop the use of percent units in metrics, we'd just want to warn people that this unit is not being propagated to the table.

jrmullaney: Following on from my last post, I've implemented the warnings I suggested. To avoid excessive warnings on the standard `--log-level`, I have it just reporting when there's "one or more" percent units used in a bundle. Users can then use the `debug` level to identify the problematic metrics.

Does that sound ok?

kherner (approved):

PR #337, opened 2025-01-15
DM-47730: fix various incorrect NoWorkFound conditions

TallJimbo:


jrmullaney (approved):
Looks fine. I've tested `_task.py` with my metric histogram plotting analysis tool (which is what instigated the need for band removal from filenames where desired), and `matadataAnalysis.py` works as I'd expect with calexpMetrics.yaml#calexpSummary. 

I've just made one query about the ValueError message, but I'm approving this (analysis tools) PR,
 
Thanks for this!

PR #343, opened 2025-01-29
DM-48655: Add pandas index to visit table on read if needed.

TallJimbo:
This exact logic was needed in another method called applyAstrometricCorrections on a different task; I'm a little concerned there might be some code duplication there, but I don't have time to dig too deeply there.


leeskelvin (approved):
LGTM.

PR #344, opened 2025-01-29
DM-47632: Move pviMetrics into visitQualityCore for calibrateImage switch.

TallJimbo:


TallJimbo (commented):
This PR looks more complicated than it is: the content of `pviMetrics` was moved as a block into `visitQualityCore`.

erykoff (approved):
This looks fine, but how did this work before with HSC (moved to calibrateImage) and ComCam (still on Calibrate)?

TallJimbo:
> This looks fine, but how did this work before with HSC (moved to calibrateImage) and ComCam (still on Calibrate)?

There is a corresponding commit in `drp_pipe` that needs to be paired with this one that swaps out which analysis pipelines are imported by the DRP pipelines.

PR #345, opened 2025-01-29
DM-45545: Fix np.Inf to np.inf for numpy2 compatibility

timj:


enourbakhsh (approved):
Thanks for fixing this. Looks good!

PR #346, opened 2025-01-30
DM-48600: Add APDB read/write timing metrics

isullivan:


pipelines/apDetectorVisitQualityCore.yaml
```diff
@@ -105,6 +105,22 @@ tasks:
         nMergedDiaSources: ct
       python: |
         from lsst.analysis.tools.atools import TaskMetadataMetricTool
+  loadDiaCatalogsCore:
+    class: lsst.analysis.tools.tasks.TaskMetadataAnalysisTask
+    config:
+      connections.inputName: loadDiaCatalogs_metadata
+      connections.outputName: loadDiaCatalogs_metadata  # Should be appended with "_metrics"
```
BrunoSanchez, line 112: Is it really the output the same as the input here?

BrunoSanchez (approved):
I think it is all good. Just a minor question about output name

pipelines/apDetectorVisitQualityCore.yaml
isullivan: No. `TaskMetadataAnalysisTask` appends `_metrics` to it's output connections, so it will use `loadDiaCatalogs_metadata_metrics`

PR #347, opened 2025-02-04
DM-48582: factor out configs that are repeated for each kind of visit/source analysis

TallJimbo:


config/refCatSourceAstrometric-standard.py
```diff
@@ -0,0 +1,16 @@
+# Configuration for standard instances of
+# sst.analysis.tools.tasks.refCatSourceAnalysis.RefCatSourceAnalysisTask
```
sr525, line 2: Missing l

config/refCatSourcePhotometric-standard.py
```diff
@@ -0,0 +1,15 @@
...
+# that includes this standard set.
+
+from lsst.analysis.tools.atools import *
+
+config.atools.photomDiffPsfSkyVisitPlot = TargetRefCatDeltaPsfSkyVisitPlot
+config.atools.photomDiffAp09SkyVisitPlot = TargetRefCatDeltaAp09SkyVisitPlot
+config.atools.photoDiffPsfScatterVisitPlot = TargetRefCatDeltaPsfScatterVisitPlot
+config.atools.photoDiffCModelScatterVisitPlot = TargetRefCatDeltaAp09ScatterVisitPlot
+
+# TODO: should there be a VisitContext set here, too?
```
sr525, line 15: The contexts are applied in each atool so probably not.

PR #352, opened 2025-02-14
DM-47181: Modify hist plot output format for Astrometry.

PFLeget:
When I was looking at plots output for Astrometry, label were out of the figure. This is fixing that and I fixed limit of plot to make easier to compare between different run.


erykoff (approved):
This looks fine, but I'm also pinging @sr525 for the default hist changes.

sr525:
If it fixes the off the edge issue I will be very happy! Have you run an coaddQuailtyCore and just checked that all of the hist plots look normal?

PFLeget:
Yes it fixe the edge issue, but I did not checked coaddQuailtyCore. Let me check.

PR #355, opened 2025-02-21
DM-49082: Make config validation for property maps more reliable in analysis_tools

enourbakhsh:


arunkannawadi (approved):
The PR is doing what the ticket says, so I approve in that sense. I'm just not convinced that using `keyCheck` is better than validating this in the `validate()` method. Sure, the advantage of `keyCheck` is that it wouldn't even let one construct an invalid config, while the latter requires an explicit `validate` call (which our pipelines do anyway). But the big disadvantage if this is checked in `validate`, it is possible to raise a useful error message that 'orientation' and 'location' are not allowed, for e.g., whereas the current implementation simply says that `orientation` is not a valid key. As a general solution, it is not possible for the error message to list out all the invalid keys with `keyCheck`, as there may be infinite of them. But on a case-by-case basis, and specifically here, we can list out which ones are invalid.

python/lsst/analysis/tools/tasks/propertyMapAnalysis.py
```diff
@@ -111,7 +108,7 @@ class PerTractPropertyMapAnalysisConfig(
         itemtype=str,
         doc="Keyword arguments to pass to the colorbar.",
```
arunkannawadi, line 109: Here, and elsewhere, I'd recommend adding in a documentation of which keys are invalid. But I also worry that it could go out of sync (and hence my comment preferring `validate` over `keyCheck`)

arunkannawadi:
Sounds good to me!

PR #357, opened 2025-02-24
DM-49077: Investigate strange PA1 metric on DP1 processing tract 5063, r band

jrmullaney:
Title says "Investigate", but this merge is the fix that addresses what was found. It ensures that the signal-to-noise cut takes place prior to group formation.

### DIFF
--- a/python/lsst/analysis/tools/atools/photometricRepeatability.py
+++ b/python/lsst/analysis/tools/atools/photometricRepeatability.py
@@ -72,6 +72,9 @@ class StellarPhotometricRepeatability(AnalysisTool):
 
         # Apply per-source selection criteria
         self.prep.selectors.bandSelector = BandSelector()
+        self.prep.selectors.snSelector = SnSelector()
+        self.prep.selectors.snSelector.fluxType = self.fluxType
+        self.prep.selectors.snSelector.threshold = 200
 
         # Compute per-group quantities
         self.process.buildActions.perGroupSn = PerGroupStatistic()


python/lsst/analysis/tools/atools/photometricRepeatability.py
```diff
@@ -72,6 +72,9 @@ def setDefaults(self):
 
         # Apply per-source selection criteria
         self.prep.selectors.bandSelector = BandSelector()
+        self.prep.selectors.snSelector = SnSelector()
+        self.prep.selectors.snSelector.fluxType = self.fluxType
```
erykoff, line 76: Some places this is `f"{self.fluxType}"` but I guess that quoting is completely redundant, isn't it?

Also, does this make the `self.process.filterActions.perGroupStdevFiltered.selectors.sn` below redundant?

jrmullaney: >> Some places this is f"{self.fluxType}" but I guess that quoting is completely redundant, isn't it?
The only reason I can think it would be necessary would be to parse, for example, {band}_psfFlux. But this wouldn't be done in the atool anyway - it'd happen in the `BasePrep`, `BaseProcess` classes in `_stages.py`. So, yes, my understanding is that it is redundant.

>> Also, does this make the self.process.filterActions.perGroupStdevFiltered.selectors.sn below redundant?
Yes, I think it does, and I was considering to removing it. But I'd want to test it to make sure, which would take more time. It was a case of do-no-harm while short on time.

erykoff (approved):

PR #358, opened 2025-02-24
DM-47320: Slightly rework how property map analysis is configured.

TallJimbo:
It's arguable a bigger overhaul (done in concert with an overhaul of how the property map creation tasks are configured) is needed here, but this smaller-scope change keeps the names the same on the current pipelines while allowing more flexibility for names in the v2 ones, without any further duplication of the list of concrete property maps.


leeskelvin (approved):
A nice revision and clean-up, LGTM.

PR #359, opened 2025-02-24
DM-49118: Update comment regarding monster naming convention.

erykoff:


fred3m (approved):
LGTM
You might want to make it `the_monster_*` just to note that there will be a date string at the end of the catalog name.

PR #369, opened 2025-03-12
DM-49274: Stop sky objects appearing in NO_DATA regions

leeskelvin:


python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -370,6 +370,7 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
     def setDefaults(self):
         self.selectWhenFalse = [
             "{band}_pixelFlags_edge",
+            "{band}_pixelFlags_nodata",
```
arunkannawadi, line 373: Do you mind adding `super().setDefaults()` above this line while you're at it?

python/lsst/analysis/tools/actions/vector/selectors.py
```diff
@@ -392,6 +393,7 @@ def __call__(self, data: KeyedData, **kwargs) -> Vector:
     def setDefaults(self):
         self.selectWhenFalse = [
             "pixelFlags_edge",
+            "pixelFlags_nodata",
```
arunkannawadi, line 396: Similarly `super().setDefaults()` here.

arunkannawadi (approved):

tests/test_diaSourceMetrics.py
```diff
@@ -30,17 +30,17 @@ class DiaSourceTest(lsst.utils.tests.TestCase):
...
 
     def test_metrics(self):
         """Test that metrics have the expected values from the test data."""
         NumDiaSources = diaAtool.NumGoodDiaSourcesMetrics()
         NumDiaSources.finalize()
         DiaSourceCount = NumDiaSources(self.data)
-        self.assertEqual(DiaSourceCount["numAllDiaSources"].quantity.value, 558)
-        self.assertEqual(DiaSourceCount["numGoodDiaSources"].quantity.value, 547)
-        self.assertEqual(DiaSourceCount["ratioGoodToAllDiaSources"].quantity.value, 547 / 558)
+        self.assertEqual(DiaSourceCount["numAllDiaSources"].quantity.value, 2994)
```
mrawls, line 41: This seems like a big change - any idea why?

leeskelvin: Yes - I didn't know the data ID that was used to generate the original dataset, so I guessed my own. This means that the output catalog has grown in size from the prior 558 entries to 2994. Undoubtedly, my data ID is not the same one that was used to generate the original test sample.

mrawls: OK that is fair! Whoever made that original dataset should have documented it better 😜 would you consider adding the script or similar you used to generate the test data for next time?

leeskelvin: I added my workflow into this Jira ticket comment: https://rubinobs.atlassian.net/browse/DM-49274?focusedCommentId=581302.

Is that a sufficient place to document this @mrawls?

python/lsst/analysis/tools/actions/vector/selectors.py
leeskelvin: Added into the `setDefaults` for `SkyObjectSelector`, `SkySourceSelector`, and `GoodDiaSourceSelector`. How does that read now @arunkannawadi?

python/lsst/analysis/tools/actions/vector/selectors.py
leeskelvin: See previous comment.

python/lsst/analysis/tools/actions/vector/selectors.py
arunkannawadi: It reads great now. Thanks, Lee

tests/test_diaSourceMetrics.py
mrawls: That is great - my initial thought was a text file or script of some kind in the new data subdirectory, but honestly having it on the Jira ticket is at least as good, thank you

PR #373, opened 2025-03-20
DM-49282: Raise NoWorkFound in AP metrics if any metadata entries are missing.

isullivan:
Only for metadata from tasks that might raise NoWorkFound during processing, and might not set all expected metadata.


Gerenjie (approved):
Looks good to me.

PR #380, opened 2025-03-28
DM-49765: Cast np.float32 to float when dispatching metrics.

ctslater:
The JSON encoder inside the requests library will throw an exception on np.float32 values (but not np.float64), so these need to be explicitly cast upstream.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
+++ b/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
@@ -35,6 +35,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, cast
 from uuid import UUID, uuid4
 
+import numpy as np
 import requests
 from lsst.daf.butler import DatasetRef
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
```diff
@@ -591,6 +592,9 @@ def _prepareBundle(
                             )
                             resultsTrimmed = True
                             continue
+                        # JSON will not serialize np.float32, must cast.
+                        if isinstance(value, np.float32):
```
timj, line 596: I think `isinstance(value, numbers.Real)` will work, which is what is done elsewhere in analysis_tools and is more general than np.float32 check.

ctslater: That also catches ints, which I was trying to avoid casting unnecessarily.

taranu (approved):
If it works, it works (sad that it's necessary though).

PR #383, opened 2025-04-03
DM-49074: Explicitly cast float32 to avoid JSON error. Backport v29.

ctslater:
Without this, single-precision metrics result in "TypeError: Object of type float32 is not JSON serializable"


mwittgen:
Might be good to give the merge titles names like DM-49074-v29

PR #384, opened 2025-04-08
DM-49219: Fix cp_verify noise tests

czwa:
Update column name.  Fix units for bias and dark plots.


erykoff (approved):
Thanks for fixing the label units here!

PR #393, opened 2025-04-21
DM-50090: Add Sersic index and size vs mag plots

taranu:


python/lsst/analysis/tools/atools/genericProduce.py
```diff
@@ -121,3 +125,47 @@ def finalize(self):
                 statAction.lowSNSelector.threshold = 10
                 statAction.fluxType = fluxType
                 setattr(self.process.calculateActions, f"stats_{plural}{suffix_y}", statAction)
+
```
sr525, line 128: This shouldn't really be in generic produce. I realise that the first bit ended up in here on a previous ticket but really genericStuff shouldn't have things in that isn't the infrastructure code.

PR #394, opened 2025-04-22
DM-50369: Use brute-force matching for large radii in astrometric association.

erykoff:
This greatly limits the memory usage and runs in a reasonable amount of time (5-10 seconds for the matching on USDF).

### DIFF
--- a/python/lsst/analysis/tools/actions/keyedData/calcDistances.py
+++ b/python/lsst/analysis/tools/actions/keyedData/calcDistances.py
@@ -54,6 +54,14 @@ class CalcRelativeDistances(KeyedDataAction):
         doc="Maximum number of pairs to use; downsample otherwise.",
         default=100_000,
     )
+    bruteForceRadiusThreshold = Field[float](
+        doc="Matching radius at which to switch to brute-force matching (deg).",
+        default=0.1,
+    )
+    bruteForceMaxMatch = Field[int](
+        doc="Number of objects to sample for brute-force matching.",
+        default=10_000,
+    )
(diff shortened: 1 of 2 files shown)


python/lsst/analysis/tools/actions/keyedData/calcDistances.py
```diff
@@ -140,12 +148,29 @@ def _compressArray(arrayIn):
...
+                sphDist(
+                    np.deg2rad(meanRa)[i1],
+                    np.deg2rad(meanDec)[i1],
+                    np.deg2rad(meanRa)[i2],
+                    np.deg2rad(meanDec)[i2],
+                ),
+            )
+        else:
+            with Matcher(meanRa, meanDec) as m:
+                idx, i1, i2, d = m.query_self(annulus[1], return_indices=True)
```
fred3m, line 171: I see what you're doing here and it is ok, but I wonder why you use a brute force search after randomly selecting objects to match. Since you've reduced the number of observations can't you just use smatch.Matcher now? I would think that it would have gotten rid of the memory issues.

erykoff: At 3 degrees it turns out that the brute force uses less memory and time than actually trying to construct the tree and create all the cross-indexes.  So the Matcher would be fine, but I think this is better and has guaranteed predictability.

erykoff: I considered also putting in a check if the radius was larger than the size of the box but deemed that over-optimization.

PR #396, opened 2025-04-23
DM-47916: Remove reference to makeWarp task

arunkannawadi:


kherner (approved):
There's a complaint about `atools.sersicSizeVsSersicMag.is_covariance: False` not being one of "false" or "true", but that's arguably beyond the scope of this ticket.

arunkannawadi:
Yup, `yamllint` has slightly different opinions based on the config and I am considering it out of scope for this ticket.

PR #398, opened 2025-04-28
DM-50533: Implement publication-style property map plots

enourbakhsh:


pipelines/DP1Paper.yaml
```diff
@@ -52,12 +52,25 @@ tasks:
...
       projectionKwargs:
         celestial: true
         gridlines: true
+        min_lon_ticklabel_delta: 0.12
+      publicationStyle: true
       # The entries in the 'atools' namespace must exactly match the dataset types.
       atools.deepCoadd_dcr_ddec_map_weighted_mean: PerTractPropertyMapTool
-      publicationStyle: true
+      atools.deepCoadd_dcr_dra_map_weighted_mean: PerTractPropertyMapTool
+      atools.deepCoadd_dcr_e1_map_weighted_mean: PerTractPropertyMapTool
```
sr525, line 63: This pipeline is supposed to be an example pipeline and doesn't need all of the property maps, just pick your favorite.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -362,35 +366,56 @@ def makePlot(
...
+                "Only the full-tract map will be plotted."
+            )
+            histColors = ["#440154"]
+            rasterized = False
+            labelpad = 14
+        else:
+            zoomFactors = plotConfig.zoomFactors
+            # Muted green for the full map, and muted red and blue for the two
+            # zoomed-in maps.
+            histColors = ["#265D40", "#8B0000", "#00008B"]
```
sr525, line 394: Can we use colours from the colourmap palette? Maybe the two ends and the middle? https://waldyrious.net/viridis-palette-generator/ can pull out the hexcodes.

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -362,35 +366,56 @@ def makePlot(
...
 
         with rc_context(rcparams):
-            fig = make_figure(figsize=(16, 16))
-            ax1 = fig.add_subplot(221)
-            ax2 = fig.add_subplot(222)
-            ax3 = fig.add_subplot(223)
-            ax4 = fig.add_subplot(224)
+            if plotConfig.publicationStyle:
+                # Make a single plot for the full tract.
+                fig = make_figure(figsize=(8, 8))
```
sr525, line 405: The figsize is set in the style file, does it not work for this plot?

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
```diff
@@ -457,42 +482,71 @@ def makePlot(
...
+                        "pad": colorBarPad,
+                        "cax": cax,
+                        **colorbarKwargs,
                     }
                 )
                 cbar.ax.tick_params(labelsize=colorbarTickLabelSize)
-                cbarText = "Full Tract" if zoomFactor == 1.0 else f"{self.prettyPrintFloat(zoomFactor)}x Zoom"
+                if plotConfig.publicationStyle:
+                    if plotInfo["unit"] not in ["dimensionless", "N/A"]:
+                        cbarText = f"{plotInfo['property']} / {plotInfo['unit']}"
```
sr525, line 543: Please follow the guidelines on units and styles.

timj: This is the Monthly Notices style where the quantities on the axes are dimensionless because they've been divided by the unit, but we had a long Slack discussion on this last year where it was decided that `/` should not be used 😄

python/lsst/analysis/tools/actions/plot/propertyMapPlot.py
enourbakhsh: Everything's tuned for an 8x8 figure. If you switch sizes, you'll need to adjust paddings, font sizes, and other things because arbitrary figsize changes without these tweaks can lead to unwanted results.

PR #400, opened 2025-05-01
DM-50656: Always rasterize the map image of property map plots

enourbakhsh:


erykoff (changes requested):
This should not be a config.  It should always be rasterized.  (Note that this just rasterizes the image, not the labels or ticks or anything).

enourbakhsh:
Removed unnecessary config options. We now always rasterize the map image, while keeping labels, axes, and other elements in scalable vector format.

erykoff (approved):

PR #409, opened 2025-07-16
DM-51537: Add limiting surface brightness metric to analysis_tools

aemerywatkins:
Analysis tool to estimate the limiting surface brightness (3sigma on 10" x 10" scales) of an image (visit or coadd) using sky objects.  This is a metric commonly used by the LSB community to report image surface brightness depth.


python/lsst/analysis/tools/atools/limitingSurfBri.py
```diff
@@ -0,0 +1,66 @@
...
+    parameterizedBand: bool = False
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.process.buildActions.x = LoadVector()
+        self.process.buildActions.y = LoadVector()
+        self.process.buildActions.x.vectorKey = "bin_mid"
+        self.process.buildActions.y.vectorKey = "hist"
+        self.process.buildActions.xerr = ConstantValue(value=0)
+        self.process.buildActions.yerr = ConstantValue(value=0)
```
sr525, line 40: Does this have to be set to 0? Can it just not be set?

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
```diff
@@ -0,0 +1,338 @@
...
+    },
+):
+    """Connections class for LimitingSurfBriTask."""
+
+    data = Input(
+        doc="Visit- or coadd-level object table",
+        name="{detectionTableName}",
+        storageClass="ArrowAstropy",
+        multiple=True,
+        dimensions=(),
```
sr525, line 62: I don't think all of these dimensions should be empty.

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
```diff
@@ -0,0 +1,338 @@
...
+    limiting_surface_brightness_table = Output(
+        name="limiting_surface_brightness_table",
+        storageClass="ArrowAstropy",
+        doc="A table containing two columns: the detector or patch IDs and the values of limiting surface "
+        "brightness derived for those detectors or patches.",
+        dimensions=(),
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
```
sr525, line 109: What is the below for?

python/lsst/analysis/tools/atools/limitingSurfBri.py
leeskelvin: If we switch to a HistPlot, this issue should resolve itself.

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
leeskelvin: I think (@aemerywatkins can correct me if wrong) that these have been left empty on purpose because the dimensions are set programatically by the user in the, e.g., `inputTableDimensions`. 

Leaving this blank may help make it clear that these are set elsewhere? We could populate them, but they'll never be used, so it would be source-code-only code, if that has any use at all?

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
sr525: Ok, so it comes from here https://community.lsst.org/t/more-support-for-configurable-pipelinetask-connections/7630, if Jim says this is the right way to do it then I believe him.

python/lsst/analysis/tools/atools/limitingSurfBri.py
aemerywatkins: Switching to HistPlot resolved this, as expected.

sr525: But it is still here and set to 0?

sr525: Is this whole file supposed to have been replaced?

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
```diff
@@ -0,0 +1,338 @@
+# This file is part of analysis_tools.
```
sr525, line 1: Should this file have been replaced by the new one?

python/lsst/analysis/tools/tasks/limitingSurfaceBrightnessAnalysis.py
```diff
@@ -0,0 +1,295 @@
...
+
+            # Isolate the sky sources per image, 9px radius apertures
+            idKey = "detector" if "detector" in dataId else "patch"
+            skyKey = "sky_source" if "detector" in dataId else "sky_object"
+            isImage = source_catalogue[idKey] == dataId[idKey]
+            isSky = source_catalogue[skyKey] > 0
+            skySources = source_catalogue[isImage & isSky][band + "ap%02dFlux" % (self.config.apertureSize)]
+
+            # Derive the clipped standard deviation of sky sources in nJy
+            nPix = np.pi * 9**2  # Number of pixels within the circular aperture
```
sr525, line 239: This version has the 9 here but I am not sure which version of this code is the correct one.

python/lsst/analysis/tools/tasks/limitingSurfBriAnalysis.py
aemerywatkins: Yes, forgot to add deleted files to the commit.  Done now.

python/lsst/analysis/tools/tasks/limitingSurfaceBrightnessAnalysis.py
aemerywatkins: Replaced this with the config value.

PR #411, opened 2025-08-07
DM-52115: Handle exceptions from all SasquatchDispatcher's network operations

kfindeisen:
This PR broadens the request error handling in `_dispatcher.py` to include cluster and topic setup (which are more likely than the upload to encounter network problems because they run first) and to include failure modes that don't return an HTTP response.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
+++ b/python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
@@ -202,14 +202,13 @@ class SasquatchDispatcher:
 
         headers = {"content-type": "application/json"}
 
-        with http_client() as session:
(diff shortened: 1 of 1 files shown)


python/lsst/analysis/tools/interfaces/datastore/_dispatcher.py
```diff
@@ -256,6 +254,9 @@ def _create_topic(self, topic_name: str) -> bool:
                     e.response.reason,
                 )
                 return False
+        except requests.RequestException:
+            log.exception("Unknown error occured creating kafka topic")
```
tcjennings, line 258: typo: s/ocurred/occurred

tcjennings (approved):

PR #420, opened 2025-08-28
DM-51791: Add per-visit photometric offsets vs. refcat metrics

jeffcarlin:
Added the following metrics summarizing the photometric residuals of a visit-level catalog vs. the refcat:
`ref_photom_offset`: median offset in mmag
`ref_photom_offset_mean`: mean offset in mmag
`ref_photom_offset_stdev`: std deviation of residuals in mmag
`ref_photom_offset_sigmaMad`: sigma_MAD of residuals in mmag
`ref_photom_offset_rms`: RMS of residuals in mmag
`ref_photom_offset_nstars`: number of stars used to calculate the residuals

### DIFF
--- a/config/refCatSourcePhotometricCore.py
+++ b/config/refCatSourcePhotometricCore.py
@@ -6,10 +6,12 @@
 # that includes this standard set.
 
 from lsst.analysis.tools.atools import *
+from lsst.analysis.tools.contexts import VisitContext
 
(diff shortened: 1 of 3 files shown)


jrmullaney (commented):
I've left some questions and suggested changes depending on the answer to those questions. The questions are about the scope of the atool.

PR #426, opened 2025-09-17
DM-52558: Fix property plot crash when there are no valid values in the map

erykoff:
This also fixes a numpy deprecation warning.


laurenam (approved):
Is this really the name you want to use for the branch?

PR #467, opened 2026-02-13
DM-53441: Create an analysis_tool to extract first-order astrometry summary stats for visit images

jrmullaney:


python/lsst/analysis/tools/tasks/simpleTableAnalysis.py
```diff
@@ -0,0 +1,126 @@
...
+from astropy.table import Column
+
+from lsst.pex.config import Field, ListField
+from lsst.pipe.base import connectionTypes as cT
+from lsst.skymap import BaseSkyMap
+
+from ..interfaces import AnalysisBaseConfig, AnalysisBaseConnections, AnalysisPipelineTask
+
+
+class SimpleTableAnalysisConnections(
```
sr525: I am still not convinced we need this and that one of the other table readers couldn't be adapted. If we do need this then the name should probably be changed before someone thinks it reads Simple AFW tables.

jrmullaney: There isn't a an existing table reader that reads-in an `ExposureCatalog` and just passes it straight to `run`. There are ones that read in an `ExposureCatalog`, but they do loads of manipulation prior to passing it to run. `self.LoadData` doen't work with an `Exposurecatalog`, so I can't use an existing  `ArrowAstropy` reader that just passes to `run`.

So, I think I have to make a new `ExposureCatalog` reader that passes straight to run anyway.

But the justification for doing this was that thare are a fair few table readers that basically do the same job - read in an `ArrowAstropy` then pass it straight to run. But they are all in different files, which could all be consolidated.

If we want, I could just make this into a `VisitSummaryAnalysisTask` for now, but I do think we should consider whether to consolidate all the tasks that basically just read and pass-on `ArrowAstropy` tables.

python/lsst/analysis/tools/atools/astrometryMetrics.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.process.buildActions.cornersep = AngularSeparation(
+            raKey_A="raCorners_0",
+            decKey_A="decCorners_0",
+            raKey_B="raCorners_2",
+            decKey_B="decCorners_2",
+            outputUnit="arcminute",
+        )
+        self.process.buildActions.ratio = DivideVector()
+        self.process.buildActions.ratio.actionA = AngularSeparation(
```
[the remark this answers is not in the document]
jrmullaney: Oh - on second thoughts, I don't think I can without introducing a `filterActions` step, which doesn't seem right for this calculation (it's not a filter).
We can't guarantee what order the `buildActions` will be done in, so we can't know for certain that `cornersep` will be built before the `DivideVector` is attempted.
So I _think_ it's better to leave as-is, but let me know otherwise.

PR #471, opened 2026-03-17
DM-54368: Make a metric that records the fraction of raws that made it into a coadd.

jrmullaney:


python/lsst/analysis/tools/atools/coaddInputCount.py
```diff
@@ -213,3 +221,45 @@ def setDefaults(self):
...
+        self.process.calculateActions.inVisitSummaryCount = SumAction(vectorKey="visitSummaryRecord")
+        self.process.calculateActions.inCoaddCount = SumAction(vectorKey="inCoadd")
+
+        self.process.calculateActions.inVisitSummaryFraction = DivideScalar()
+        self.process.calculateActions.inVisitSummaryFraction.actionA = SumAction(
+            vectorKey="visitSummaryRecord"
+        )
+        self.process.calculateActions.inVisitSummaryFraction.actionB = CountAction(vectorKey="inCoadd")
+
+        self.process.calculateActions.overlapFraction = DivideScalar()
```
taranu: You can just pass `actionA=SumAction(...)` and `actionB=CountAction(...)` kwargs to the constructor here, and same below.

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
```diff
@@ -0,0 +1,229 @@
...
+        doc="""Table summarising the visit images.
+            "Contains image corners that are used to
+            "refine the coverage.""",
+        name="ccdVisitTable",
+        storageClass="ArrowAstropy",
+        dimensions=("instrument",),
+        deferLoad=True,
+        multiple=False,
+    )
+    # visitSummary isn't used, but a per-visit data input
```
taranu: Why isn't this just in the doc kwarg?

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
```diff
@@ -0,0 +1,229 @@
...
+    # substantially reduces the time to build quantum graph.
+    visitSummary = cT.Input(
+        doc="""Table summarising the visit images.
```
taranu: This is the same as for visitTable; they can't both be right.

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
```diff
@@ -0,0 +1,229 @@
...
+            "lrcdec",
+        ]
+        imageCorners = self.loadData(visitTableHandle, columns)
+
+        # Group raws by visit so the imageCorners table can be reduced by visit
+        rawsByVisit = defaultdict(list)
+        for rawRef in inputRefs.raw:
+            rawsByVisit[rawRef.dataId["exposure"]].append(rawRef)
+
+        visits = []
```
taranu: Everything from here to the `self.run` call should be put in an overloaded `run` method that then calls `super().run(...)`. Or maybe even have a separate `makeData` method.

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
```diff
@@ -0,0 +1,229 @@
...
+                visits += [visit] * len(rawRefs)
+                detectors += [rawRef.dataId["detector"] for rawRef in rawRefs]
+                inCoadd_col += [False] * len(rawRefs)
+                visitSummaryRecord += [False] * len(rawRefs)
+                overlapsWithPatch += [False] * len(rawRefs)
+                continue
+
+            for rawRef in rawRefs:
+                detector = rawRef.dataId["detector"]
+                visits.append(visit)
```
taranu: I imagine this task is IO limited but appending to a list is very slow, so if you can preallocate these lists and assign by index that should reduce runtime.

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
```diff
@@ -0,0 +1,229 @@
...
+                    # If there's no visit summary, record and move on:
+                    visitSummaryRecord.append(False)
+                    overlapsWithPatch.append(False)
+                    continue
+
+                # If we have got this far, then there is a row for this image.
+                # Record whether the calibrated image covers the patch:
+                visitSummaryRecord.append(True)
+
+                raCorners = np.array(
```
taranu, line 228: This is also probably not a performance bottleneck, but you could do something like:
```
detCorners = [None]*4
visitRows = visitSummary[rowMask][0]
failed = False
for idx_det, (ra_col, dec_col) in enumerate(("llcra", "lldec"), ("ulcra", "ulcdec"), ("urcra", "urcdec"), ("lrcra", "lrcdec")):
    if math.isfinite(ra := ra_col) and math.isfinite(dec := dec_col):
        detCorners[idx_det] = SpherePoint(ra, dec, degrees).getVector()
    else:
         failed = True
         break
if not failed:
    ...
```

If you're going to use SpherePoint, degrees, convexHull etc. in nested loops it's better to do `from ... import` and save a bunch of repeated getattrs on modules.

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
```diff
@@ -0,0 +1,229 @@
...
+                    ]
+                )
+                decCorners = np.array(
+                    [
+                        visitSummary[rowMask][corner].value[0]
+                        for corner in ["llcdec", "ulcdec", "urcdec", "lrcdec"]
+                    ]
+                )
+
+                if np.all(np.isfinite(raCorners)) and np.all(np.isfinite(decCorners)):
```
taranu, line 243: When/why are they not finite?

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
```diff
@@ -0,0 +1,229 @@
...
+                    overlapsWithPatch.append(False)
+                    continue
+
+                # If we have got this far, then there is a row for this image.
+                # Record whether the calibrated image covers the patch:
+                visitSummaryRecord.append(True)
+
+                raCorners = np.array(
+                    [
+                        visitSummary[rowMask][corner].value[0]
```
taranu, line 230: Does this work as intended if `len(visitSummary[rowMask]) > 1`? I can't quite tell by eye.

taranu (commented):
This mostly looks fine, but if this is a memory-intensive task it really needs a unit test more than usual.

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
jrmullaney: I didn't put it in an overloaded run method, but I did move it to a separate method called "makeData".

python/lsst/analysis/tools/tasks/coaddInputAnalysisTask.py
jrmullaney: I now index, since we know how big the array will be from the outset. But I don't think it'll have much impact on run time (as you say, I/O limited).

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
jrmullaney: I didn't implement this, as it'd be a negligible speed-up and I think the original is far easier to read.

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
jrmullaney: I think it's a good to code defensively here. I could imagine some kind of catastrophic WCS fail that would NaNs for the corners. I don't think there's any harm here.

python/lsst/analysis/tools/tasks/coaddInputAnalysis.py
jrmullaney: Now this is definietely something that shouldn't happen! There should never be more than one entry in the visitTable for a (visit,detector). So I've added a RuntimeError if there's ever more than one.

PR #473, opened 2026-03-20
DM-54446: Allow histPlot to plot a variable number of panels per row

leeskelvin:


python/lsst/analysis/tools/actions/plot/histPlot.py
```diff
@@ -224,6 +224,16 @@ class HistPlot(PlotAction):
         "A number of custom color maps are also defined: `newtab10`, `bright`, `vibrant`.",
         default="rubin",
     )
+    panelsPerRow = Field[int](
+        doc="Maximum number of histogram panels to place in each row. Set to 1 to stack panels vertically.",
+        default=2,
+    )
```
sr525, line 230: Does this default change the behaviour for any other plots?

leeskelvin: No, this default preserves existing behaviour in other HistPlot outputs.

PR #478, opened 2026-04-15
DM-54270: Improve object and add ellipticity plots

taranu:


python/lsst/analysis/tools/atools/ellipticityMagnitude.py
```diff
@@ -0,0 +1,66 @@
...
+    def setDefaults(self):
+        super().setDefaults()
+        self.produce.plot.yLims = (0.0, 1.0)
+        self.produce.plot.legendLocation = "upper right"
+
+
+class ExponentialEllipticityMagnitudePlot(EllipticityMagnitudePlotBase):
+    def setDefaults(self):
+        super().setDefaults()
+        self.action_column = SubtractVector(
```
jrmullaney: Shouldn't this be `action_vector`? I get an AttributeError when I try to run this:
```
AttributeError: lsst.analysis.tools.atools.ellipticityMagnitude.ExponentialEllipticityMagnitudePlot has no attribute action_column
```
Looking at the parent class, I suspect it should be `action_vector`.

python/lsst/analysis/tools/atools/actionMagnitudeScatterPlot.py
```diff
@@ -0,0 +1,94 @@
...
+        self.prep.selectors.flagSelector = VisitPlotFlagSelector()
+
+    def finalize(self):
+        # A lazy check for whether finalize has already been called
+        classes = self.get_classes()
+        if hasattr(self.process.filterActions, self.get_name_attr_values(classes[0])):
+            return
+        super().finalize()
+
+        attr_column = f"get_{self.key_y}"
```
jrmullaney, line 63: Maybe add a validation in case a user forgets to define `key_y`? Just so it fails nicely.

python/lsst/analysis/tools/atools/ellipticityMagnitude.py
```diff
@@ -0,0 +1,66 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from ..actions.vector import LoadVector
+from ..actions.vector.mathActions import ConstantValue, DivideVector, SubtractVector
+from .actionMagnitudeScatterPlot import ActionMagnitudeScatterPlot
+
+__all__ = (
+    "EllipticityMagnitudePlotBase",
```
jrmullaney, line 27: Will `EllipticityMagnitudePlotBase` ever be used standalone, rather than being a base class? If not, does it need to be in `_all_`?

python/lsst/analysis/tools/atools/ellipticityMagnitude.py
```diff
@@ -0,0 +1,66 @@
...
+            ),
+        )
+        self.key_y = "axrat_exponential"
+        self.produce.plot.yAxisLabel = "Exponential ellipticity"
+
+
+class SersicEllipticityMagnitudePlot(EllipticityMagnitudePlotBase):
+    def setDefaults(self):
+        super().setDefaults()
+        self.action_column = SubtractVector(
```
jrmullaney: And the same again here, I think.

jrmullaney (changes requested):
See potential bug report below regarding action_column/action_vector.
The other ones are minor, though.
Outside the code - did you see that the position of the legend has gone all screwy in `object_exponentialEllipticityVsExponentialMag_ScatterPlotWithTwoHists`?
It's normally nicely contained at the bottom of the plot, but in this plot it has jumped to the top and now overflows the plot boundary on the left side. Can you fix that (it's not covered by the files that you've changed, so there's no '+' for me to click on to add a comment)?

python/lsst/analysis/tools/atools/ellipticityMagnitude.py
taranu: Yeah, this was from an annoying rebase conflict that I neglected to push the fix for before making the PR.

python/lsst/analysis/tools/atools/actionMagnitudeScatterPlot.py
taranu: Hmm. I was going to say this is not necessary because tools used in tasks have `validate` called before `finalize`, and this field is not optional so it will fail. However, you can instantiate and finalize tools outside of tasks. I suppose I'll add a call to `validate` here then ping @natelust because I can't remember if we discussed this before.

It's not a problem to call validate multiple times but I'm not sure if we should make a habit of it either.

python/lsst/analysis/tools/atools/ellipticityMagnitude.py
taranu: It _can_ be used standalone, either on its own (by passing in `action_vector` and the other necessary kwargs) or in a pipeline (with overrides in the yaml). I wouldn't recommend it, though, which is why I tacked on `Base`.

Incidentally, you may have noticed that atools can't be real ABCs due to a metaclass conflict, so I don't think I can do much more than have the suffix there. I will add a class docstring, though.

jrmullaney (approved):

PR #479, opened 2026-04-27
DM-54367: Metrics of coadd masked pixel numbers at patch, tract, and sky-level

jrmullaney:


config/wholeTractMaskFractionMetrics.py
```diff
@@ -0,0 +1,64 @@
...
+        action = agg_cls()
+        action.vectorKey = f"{plane}_fraction"
+        setattr(
+            config.atools.tractMaskFractions.process.calculateActions,
+            f"{plane}_fraction_{agg_name}",
+            action,
+        )
+
+        # Mask fractions of non-NO_DATA (i.e., "valid") patch regions:
+        if plane != "NO_DATA":
```
cmsaunders, line 29: Are you deliberately changing what you set at line 43, or could you do this all in one block above?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -103,6 +104,26 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+    weights."""
+
+    vectorKey = Field[str]("Key of Vector of values to compute the weighted mean of.")
+    weightsKey = Field[str]("Key of Vector of weights.")
+
+    def getInputSchema(self) -> KeyedDataSchema:
+        return ((self.vectorKey, Vector), (self.weightsKey, Vector))
+
+    def __call__(self, data: KeyedData, **kwargs) -> Scalar:
+        values = _dataToArray(data[self.vectorKey.format(**kwargs)])
```
cmsaunders: I know in the use-case here, the weights contain the masking, but I could imagine other scenarios where you might have a separate mask. Do you think it's worth included a mask here, like in `MeanAction`?

python/lsst/analysis/tools/actions/scalar/scalarActions.py
```diff
@@ -103,6 +104,26 @@ def __call__(self, data: KeyedData, **kwargs) -> Scalar:
...
+        weights = _dataToArray(data[self.weightsKey.format(**kwargs)])
+        valid = ~np.isnan(values)
+        total_weight = np.sum(weights[valid])
+        if total_weight == 0:
+            return cast(Scalar, np.nan)
+        return cast(Scalar, np.sum(values[valid] * weights[valid]) / total_weight)
```
cmsaunders: Consider `np.average()`.

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
```diff
@@ -96,18 +99,23 @@ def __init__(self, *, config=None):
                 multiple=self.data.multiple,
             )
 
+        if config.outputDimensions:
+            self.dimensions.update(frozenset(sorted(config.outputDimensions)))
```
cmsaunders, line 106: If `outputDimensions` is supposed override `dimensions`, do you need to remove the existing dimensions before adding the new ones?

config/wholeTractMaskFractionMetrics.py
```diff
@@ -0,0 +1,64 @@
...
+    "SAT",
+    "SPIKE",
+    "STREAK",
+    "SUSPECT",
+    "UNMASKEDNAN",
+    "VIGNETTED",
+]
+
+# Add NO_DATA via union in case it's already included above.
+all_planes = set(config.atools.tractMaskFractions.maskPlanes) | {"NO_DATA"}
```
cmsaunders: Is this necessary, given that you also always include NO_DATA in `_collectAllPlanes` in `WholeTractMaskFractionAnalysisTask`?

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
```diff
@@ -192,6 +200,95 @@ def parsePlotInfo(
...
+            no_data_bit = mask.getPlaneBitMask("NO_DATA")
+            not_no_data = (arr & no_data_bit) == 0
+
+            total = arr.size
+            n_valid = int(np.sum(not_no_data))
+            valid_pixel_counts.append(n_valid)
+            for plane in all_planes:
+                try:
+                    bit = mask.getPlaneBitMask(plane)
+                except Exception:
```
cmsaunders: Can you make this a specific error? I don't think you want to just continue on any exception.

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
```diff
@@ -192,6 +200,95 @@ def parsePlotInfo(
...
+                    continue
+                flagged = (arr & bit) != 0
+                fractions.setdefault(f"{plane}_fraction", []).append(float(np.sum(flagged)) / total)
```
cmsaunders: I think if you set `fractions = defaultdict(list)` at line 262, then made this line 
`fractions[f"{plane}_fraction"].append(float(np.sum(flagged)) / total)`
that would be more readable.

tests/test_wholeTractMaskFraction.py
```diff
@@ -0,0 +1,250 @@
...
+
+class _MockExposure:
+    def __init__(self, array, plane_bits=_PLANE_BITS):
+        self._mask = _MockMask(array, plane_bits)
+
+    def getMask(self):
+        return self._mask
+
+
+class _MockHandle:
```
cmsaunders: You could use `lsst.pipe.base.InMemoryDatasetHandle` instead of doing this.

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
```diff
@@ -192,6 +200,95 @@ def parsePlotInfo(
...
+        all_planes : `set` [`str`]
+        """
+        mask_planes = set()
+        for fieldName in self.config.atools.fieldNames:
+            atool = getattr(self.config.atools, fieldName)
+            if hasattr(atool, "maskPlanes"):
+                mask_planes.update(atool.maskPlanes)
+        return mask_planes | {"NO_DATA"}
+
+    def _computeKeyedData(self, handles, all_planes):
```
cmsaunders: A more specific function name, like `_computeMaskFractions()`, would be more helpful. `_computeKeyedData()` could be doing anything.

tests/test_wholeTractMaskFraction.py
```diff
@@ -0,0 +1,250 @@
...
+        """SAT and CR fractions should be computed independently."""
+        arr = np.zeros((4, 4), dtype=np.int32)
+        arr[:1, :] = _PLANE_BITS["SAT"]  # 4/16 = 0.25 SAT
+        arr[1:2, :] = _PLANE_BITS["CR"]  # 4/16 = 0.25 CR
+        result = self._compute([_MockHandle(arr)])
+        self.assertAlmostEqual(result["SAT_fraction"][0], 0.25)
+        self.assertAlmostEqual(result["CR_fraction"][0], 0.25)
+
+
+class MyMemoryTestCase(lsst.utils.tests.MemoryTestCase):
```
cmsaunders, line 223: What does this do? It's new to me.

cmsaunders (approved):
Looks pretty good. I have some comments below, none major. I put some bigger questions that are more about the approach for this ticket on Jira.

config/wholeTractMaskFractionMetrics.py
jrmullaney: Yes: for each mask plane, I want the flagged fraction of all pixels, and the flagged fraction of all that are not `NO_DATA`. But I put the latter inside an `if`i conditional as there's no point calculating the fraction of not `NO_DATA` pixels for the `NO_DATA` mask plane (it would always be 0). Or have I misunderstood the question?
Remember - the `config.atools.tractMaskFractions.process.calculateActions` get build up one by one, so line 61 is not overriding line 43.

config/wholeTractMaskFractionMetrics.py
jrmullaney: No. It's not necessary, it was just belt-and-braces. I'll remove this line.

python/lsst/analysis/tools/actions/scalar/scalarActions.py
jrmullaney: Weird - I thought I'd added that. The test assumes the mask is applied. I guess it must have got lost in a `git rebase -i`.

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
jrmullaney: Good cetch. Yes. I think this is normally used when self.dimensions={}, but in this case it's already assigned, so we'll need to clear it.

tests/test_wholeTractMaskFraction.py
jrmullaney: I'm not sure. It's in all the analysis_tools test suites, so I just assumed it was needed. Sorry!

config/wholeTractMaskFractionMetrics.py
cmsaunders: Yeah, I'm not 100% sure I remember exactly what I was thinking, but I think I forgot the calculateActions could get built up.

tests/test_wholeTractMaskFraction.py
cmsaunders: I see  some other places in the stack where there's a comment saying that this checks for memory leaks.

python/lsst/analysis/tools/tasks/wholeTractImageAnalysis.py
```diff
@@ -192,6 +203,95 @@ def parsePlotInfo(
...
+            Arrays of per-patch fractions, keyed by ``"{plane}_fraction"``
+            and ``"{plane}_data_fraction"``, plus ``"valid_pixel_count"``.
+        """
+        fractions: defaultdict[str, list[float]] = defaultdict(list)
+        valid_pixel_counts: list[int] = []
+
+        for handle in handles:
+            exp = handle.get()
+            mask = exp.getMask()
+            arr = mask.array
```
cmsaunders, line 283: How about `valid_data`? `not_no_data` makes my head hurt.

(Did this go through in my initial review? Github is showing it as a draft comment for me, and it should be on line 274.)

PR #480, opened 2026-05-04
DM-54862: Enable ExtendednessTool-like behavior

arunkannawadi:
for backward compatibility. This introduces the deprecation that was missed in #478 .

### DIFF
--- a/python/lsst/analysis/tools/atools/sizeMagnitude.py
+++ b/python/lsst/analysis/tools/atools/sizeMagnitude.py
@@ -19,12 +19,56 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import logging
+
+from deprecated.sphinx import deprecated
+
 from ..actions.vector import CoaddPlotFlagSelector, VisitPlotFlagSelector
 from .genericBuild import SizeTool
 from .genericProduce import MagnitudeScatterPlot
 
+logging.basicConfig()
+_LOG = logging.getLogger(__name__)
+_LOG.setLevel(logging.WARNING)
+
 
 class SizeMagnitudePlot(SizeTool, MagnitudeScatterPlot):
+
+    _parameterizedBand: bool = True
+
+    # TODO: Remove the getter and setting in DM-54864.
+    # Instead, just convert _parameterizedBand to parameterizedBand.
+    @property
+    def parameterizedBand(self) -> bool:
+        return self._parameterizedBand
+
+    @parameterizedBand.setter
+    def parameterizedBand(self, value: bool) -> None:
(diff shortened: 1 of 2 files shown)


python/lsst/analysis/tools/atools/sizeMagnitude.py
```diff
@@ -19,12 +19,57 @@
...
+
+    @parametrizedBand.setter
+    def parameterizedBand(self, value: bool) -> None:
+        _LOG.info(
+            "Setting 'parameterizedBand' of a `SizeMagnitudePlot' is a no-op. "
+            "If you see this message, it is likely because you are trying "
+            "to read older version of configs with newer versions of the "
+            "LSST Science Pipelines."
+        )
+
```
taranu, line 71:
```suggestion
    parameterizedBand = Field[bool](
        default=True,
        deprecated="This Config field is no longer used and will revert to"
                   " AnalysisTool.parameterizedBand after v31.",
        doc="Does this AnalysisTool support band as a name parameter",
    )
```
The existing version will work since we never override `parameterizedBand` to be `False`, but making it impossible to override is not necessary afaik.

python/lsst/analysis/tools/atools/sizeMagnitude.py
```diff
@@ -19,12 +19,57 @@
...
+_LOG.setLevel(logging.WARNING)
+
 
 class SizeMagnitudePlot(SizeTool, MagnitudeScatterPlot):
+
+    # TODO: Remove the getter and setter in DM-54864.
+    @property
+    @deprecated(reason="This is no longer used.", version="v31")
+    def extendedness(self) -> None:
+        """A config-like attribute for backward compatability.
```
taranu: compatibility

taranu (commented):
It may be better to re-introduce `parametrizedBand` as a Field for now.

Either way, my main comment is that I missed doing the same thing for `ColumnMagnitudeScatterPlot`, also in atools. That's the last remaining "use" of `ExtendednessTool`, so it could be deprecated after that. Since you've already made the removal, I figure it would be better to at least get rid of all uses of it on this ticket. I can add the commit to do that if you like.

Arguably deprecating `ExtendednessTool` itself should be another ticket and/or trivial RFC, since it is a public interface - up to you, I suppose.

arunkannawadi:
@taranu I've made the changes in `ColumnMagnitudeScatterPlot`. I decided against reintroducing these as `Field` instances again: as config fields, they'd be persisted, regardless of their deprecation status, and given that they were removed, reintroducing them seems a step backwards to me. They are now as close as they can be to what they would become after removal - a class attribute.

As for deprecating `ExtendednessTool` itself, I'll leave that up to you.

taranu (approved):

PR #481, opened 2026-06-01
Add Authorization header to SasquatchDispatcher

afausti:
This is required to use a token for authentication.


hsinfang (approved):
LGTM

natelust (approved):
I always knew putting that token field in there would come in handy.

PR #483, opened 2026-06-29
DM-55241: Explicitly turn off multiprocessing in treecorr for rho stats.

erykoff:


python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
```diff
@@ -49,6 +49,9 @@
 
 _LOG = logging.getLogger(__name__)
 
+# We need to explicitly turn off multiprocessing in treecorr.
+treecorr.set_max_omp_threads(1)
```
jrmullaney, line 53: There'll be a good reason, I'm sure, but I just want to check that we're not being too conservative here. Why do we want to switch off all multithreading, rather than just keeping the number small but non-unity?

jrmullaney (approved):

python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py
yalsayyad: We need to be able to control the number of cores/threads that we use and we do that in the middleware level not the piplines payload level.

timj: The middleware passes the allowed number of threads into the runQuantum method so in theory that value could be passed into the treecorr configuration.

erykoff: In the case of analysis tools, we could do that.  For the treegp code we cannot.  Right now I wanted to get this unblocked.  Once treecorr is fixed to obey the OMP env vars I think we could do that ... I'll file a follow-up ticket.

At the same time, I think all implicit multithreading is evil, and think we should SHUT IT ALL DOWN and expand if it makes sense.

PR #487, opened 2026-08-06
DM-55750: wholeSkyPlot updates

sr525:


python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -403,69 +402,68 @@ def makePlot(
...
         else:
             xlim, ylim = self.xLimits, self.yLimits
-        ax.set_xlim(xlim)
-        ax.set_ylim(ylim)
-        ax.set_xlabel(self.xAxisLabel)
-        ax.set_ylabel(self.yAxisLabel)
-        ax.invert_xaxis()
+        # This doesn't work but I would like it to.
+        # sp.ax.set_xlim(xlim)
+        # sp.ax.set_ylim(ylim)
```
erykoff: You need to do `sp.set_extent([xlim[0], xlim[1], ylim[0], ylim[1]])`

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -314,8 +313,51 @@ def makePlot(
...
+        # Create patches using the corners of each tract.
+        tracts = []
+        ras = []
+        decs = []
+        mid_ras = []
+        mid_decs = []
+        allPatches = []
+        norm = Normalize(vmin=vmin, vmax=vmax, clip=True)
+        for i, tract in enumerate(data["tract"]):
+            corners = getTractCorners(skymap, tract)
```
erykoff, line 329: Unfortunately, this does not return the correct values.  So either you need to fix it or get the correct tract corners yourself here.  It may be that somebody wants those values for the (outer) tract corners, but they are not what you want to plot here.

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -314,8 +313,51 @@ def makePlot(
...
+                tractRas.append(ra)
+                tractDecs.append(dec)
+            minRa = np.min(tractRas)
+            maxRa = np.max(tractRas)
+            minDec = np.min(tractDecs)
+            maxDec = np.max(tractDecs)
+            width = maxRa - minRa
+            height = maxDec - minDec
+            if width > 10:
+                width = minRa + 360 - maxRa
```
erykoff, line 342: I'm trying to think of a better way of doing this.  The problem is that if you could specify the corners or the rectangle (rather than pos and width/height) it would work fine in skyproj which understands spheres.  But with the Rectangle interface you might be stuck with this.

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -265,41 +274,31 @@ def makePlot(
...
-        # Add colored patches showing tract metric values.
-        patchCollection = PatchCollection(patches, cmap=colorMap, norm=norm)
-        ax.add_collection(patchCollection)
+        if not self.publicationStyle:
+            ax = fig.add_subplot(gs[:3])
+            sp = skyproj.McBrydeSkyproj(ax=ax)
+        else:
+            ax = fig.add_subplot(111)
+            sp = skyproj.McBrydeSkyproj(ax=ax, extent=[-180, 180, -90, 30])
+        proj = sp.ax.projection
```
erykoff, line 296: As I have learned the hard way debugging skyproj (now merged to main, not released yet and not suitable for this PR right now), you have to do `proj = copy.copy(sp.ax.projection)`.  (And `import copy` on top).  Otherwise there is a race condition and you may not get rectangles.

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -330,66 +372,69 @@ def makePlot(
...
-                outlierPatchCollection = PatchCollection(
-                    outlierPatches,
-                    cmap=colorMap,
-                    norm=norm,
-                    facecolors="none",
-                    edgecolors=outlierColor,
-                    linewidths=0.5,
-                    zorder=100,
+            for i, tract in enumerate(data["tract"][outlierInds]):
+                corners = getTractCorners(skymap, tract)
```
erykoff, line 378: Here's that bad getTractCorners again.

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -330,66 +372,69 @@ def makePlot(
...
-                    linestyles="dotted",
-                    linewidths=0.5,
-                    zorder=100,
-                )
-                ax.add_collection(nanPatchCollection)
-                # Add legend information.
-                nanPatch = Patch(
-                    facecolor="white",
+            for i, tract in enumerate(data["tract"][nanInds]):
+                corners = getTractCorners(skymap, tract)
```
erykoff, line 405: It keeps on coming!

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -330,66 +372,69 @@ def makePlot(
...
-                    outlierPatches.append(patches[ind])
-                outlierPatchCollection = PatchCollection(
-                    outlierPatches,
-                    cmap=colorMap,
-                    norm=norm,
-                    facecolors="none",
-                    edgecolors=outlierColor,
-                    linewidths=0.5,
-                    zorder=100,
+            for i, tract in enumerate(data["tract"][outlierInds]):
```
erykoff, line 377: Is all this duplicated code necessary?  Can't you loop over all tracts and check as you add each one if it's an inlier, an outlier, or a nan-liar?

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -265,41 +275,34 @@ def makePlot(
...
-        # Add colored patches showing tract metric values.
-        patchCollection = PatchCollection(patches, cmap=colorMap, norm=norm)
-        ax.add_collection(patchCollection)
+        gs = gridspec.GridSpec(1, 20)
+        if not self.publicationStyle:
+            ax = fig.add_subplot(gs[3:15])
+            sp = skyproj.McBrydeSkyproj(ax=ax)
+        else:
+            ax = fig.add_subplot(111)
+            sp = skyproj.McBrydeSkyproj(ax=ax, extent=[-180, 180, -90, 30])
```
erykoff: So this is specifically and always the LSST region?

python/lsst/analysis/tools/actions/plot/wholeSkyPlot.py
```diff
@@ -265,41 +275,34 @@ def makePlot(
...
+
+        # Eli says this step is necessary and he knows more
+        # than me about most things but especially skyproj.
```
erykoff: You do not need this comment here.  But you can say that this copy is necessary to avoid a possible race condition when drawing figures in skyproj, and has been fixed in an upcoming version of skyproj.

erykoff: The other thing is that the newest version of skyproj allows you to draw the patch directly ... but this isn't in our env yet.

python/lsst/analysis/tools/utils.py
```diff
@@ -54,8 +54,15 @@ def getTractCorners(skymap, tractId):
     -----
     Corners are returned in degrees and wrapped in ra.
     """
-    tractCorners = skymap[tractId].getVertexList()
-    corners = _wrapRa([(corner.getRa().asDegrees(), corner.getDec().asDegrees()) for corner in tractCorners])
+    # Get bounding box is needed if the skymap isn't rings
+    # The discrete skymaps don't return a box from
+    # getInnerSkyRegion
+    tractBox = skymap[tractId].getInnerSkyRegion().getBoundingBox()
```
erykoff: If the inner sky region is a Box then is this final thing a no-op?

erykoff (approved):
A couple of minor comments.  This looks great, and the plots look great!

PR #489, opened 2026-08-31
DM-55998: Hard code Metrics Bundle support into datastore

timj:
Rather than relying on this package to provide an overrides configuration that specifies only metrics bundles can be accepted, hard code this knowledge into the package and allow external configuration to add to the list of accepted duck-typed storage classes.

Removes the local DAF_BUTLER_CONFIG_PATH setting and standardizes the override search name to match that expected from other datastores.

### DIFF
--- a/python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
+++ b/python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
@@ -30,6 +30,7 @@ from typing import TYPE_CHECKING, Any, ClassVar
 
 from lsst.daf.butler import DatasetRef, DatasetTypeNotSupportedError, StorageClass
 from lsst.daf.butler.datastore import DatasetRefURIs, DatastoreConfig, DatastoreOpaqueTable
+from lsst.daf.butler.datastore.constraints import Constraints, ConstraintsConfig
 from lsst.daf.butler.datastore.generic_base import GenericBaseDatastore
 from lsst.daf.butler.datastore.record_data import DatastoreRecordData
 from lsst.daf.butler.registry.interfaces import DatastoreRegistryBridge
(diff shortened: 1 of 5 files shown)


python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -64,9 +65,18 @@ class SasquatchDatastore(GenericBaseDatastore):
...
+    searches when building butler configurations. Relative to each entry of
+    ``DAF_BUTLER_CONFIG_PATH``.
+    """
+
+    builtinConstraints: ClassVar[Mapping[str, list[str]]] = {"accept": ["MetricMeasurementBundle"]}
+    """Constraints the datastore imposes on itself.
+
+    The dispatcher can only serialize metric bundles, so these types are
+    always accepted. Configuration can add further accepted types but can
+    not remove these.
```
andy-slac: If one can add additional types would that mean that dispatcher could handle other types? In which case someone may want to handle only those types and not MetricMeasurementBundle?

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
```diff
@@ -107,6 +117,15 @@ def __init__(
...
         super().__init__(config, bridgeManager)
 
+        # Datastore.__init__ builds constraints purely from configuration,
+        # which leaves the datastore accepting everything when no override
+        # file is found. Fold in the types this datastore can actually
+        # handle so that a missing or incomplete config can not widen them.
+        self.constraints = Constraints(
+            self._merged_constraints(self.config.get("constraints")),
+            universe=bridgeManager.universe,
+        )
```
andy-slac: Feels a bit odd first initialize `self.constraints` in the base class and then reinitialize here. What if base class `__init__` does something extra that depends on `constraints` (I know it does not do it now)? Would it be cleaner to update config before passing it to the base class?

andy-slac (approved):
Looks good, with a couple of questions.

python/lsst/analysis/tools/interfaces/datastore/_sasquatchDatastore.py
timj: I have been worrying about this. The actual requirement is that this datastore can only accept python types that are duck-typed like MetricMeasurementBundles. The storage class hard-coding is a proxy for that. Maybe a couple of hasattr calls would work just as well to decline the `put` without having to mess with external constraints.

timj: Going back in time we should have defined a Protocol for measurement bundles and then said that we accept based on that Protocol being defined for the python type.

timj:
@andy-slac rejigged to use a protocol test directly instead of hard-coding a storage class.

PR #490, opened 2026-09-03
DM-55321: Add a config parameter to analysis_tools base to switch on/off plot generation

jrmullaney:


pipelines/visitQualityCore.yaml
```diff
@@ -38,66 +38,19 @@ tasks:
...
-        from lsst.analysis.tools.atools import TaskMetadataMetricTool
+      connections.outputName: calibrateImage_metadata_aggregate
+      inputDataDimensions: ["instrument", "visit", "detector"]
+      outputDimensions: ["instrument", "visit"]
+      file: $ANALYSIS_TOOLS_DIR/config/aggregateCalibrateImageMetadata.py
+  makeMetricTableAggregateCalibrateImageMetadata:
+    class: lsst.analysis.tools.tasks.MakeMetricTableTask
+    config:
+      connections.metricBundleName: "calibrateImage_metadata_aggregate_metrics"
+      connections.outputTableName: "calibrateImage_metadata_aggregate_metricsTable"
```
sr525, line 53: I think some recent update changed the default names to metrics_table rather than metricsTable but I might be wrong.

tests/test_aggregatedMetadata.py
```diff
@@ -0,0 +1,300 @@
...
+    Returns
+    -------
+    task : `AggregatedTaskMetadataAnalysisTask`
+        A configured task.
+    """
+    config = AggregatedTaskMetadataAnalysisConfig()
+    config.connections.inputName = "myTask_metadata"
+    config.connections.outputName = "myTask_metadata_agg"
+    config.atools.myTool = AggregatedTaskMetadataMetricTool
+    config.atools.myTool.taskName = "myTask"
```
sr525, line 93: I'm guessing this all gets overwritten? Perhaps a different placeholder name.

sr525: Somehow missed that this was in a test, I would prefer a different name but I don't think it really matters.

tests/test_aggregatedMetadata.py
```diff
@@ -0,0 +1,300 @@
...
+
+class TestAggregatedTaskMetadataAnalysisTask(TestCase):
+    """Tests for AggregatedTaskMetadataAnalysisTask._collectData."""
+
+    def testEmptyMetadataRaisesNoWorkFound(self):
+        """All inputs returning empty metadata should raise NoWorkFound."""
+        task = _make_task({"nStars": "ct"})
+        handles = _make_handles([{}, {}, {}])
+        with self.assertRaises(UpstreamFailureNoWorkFound):
+            task._collectData(handles, "myTask")
```
sr525, line 238: Is there a count of how many that don't have any/all metrics?

